Issue #36, opened 2023-10-31
[epic] Improve logging


mastoffel:
Logging can be improved, e.g. by:
- [x] #127
- [x] #126
- [x] #68 
- [x] #151

mastoffel:
Closing as all boxes have been ticked.

PR #106, opened 2024-01-12
Refactor hyperparam search

mastoffel:
rewrite of the hyperparameter search part to be more concise and functional instead of a class


autoemulate/hyperparam_searching.py
```diff
@@ -0,0 +1,175 @@
...
+        searcher = BayesSearchCV(
+            model,
+            param_grid,
+            n_iter=niter,
+            cv=cv,
+            n_jobs=n_jobs,
+            refit=True,
+        )
+    elif search_type == "grid":
+        raise NotImplementedError("Grid search not available yet.")
```
edaub, line 81: Probably want an ``else`` here to raise a more useful error message if ``search_type`` isn't an allowed value. You may also want to something like converting the string to lower case to be a bit more tolerant of the range of inputs you can accept. You may also want to permit passing a sklearn CV instance directly if they want to configure it themselves.

autoemulate/hyperparam_searching.py
```diff
@@ -0,0 +1,175 @@
...
+    -------
+    param_grid : dict
+    """
+    if type(param_grid) != dict:
+        raise TypeError("param_grid must be a dictionary")
+    for key, value in param_grid.items():
+        if type(key) != str:
+            raise TypeError("param_grid keys must be strings")
+        if type(value) != list:
+            raise TypeError("param_grid values must be lists")
```
edaub, line 153: Is there a logic behind how you raise some errors directly and log others? It looked like logging seems to focus on the fitting errors, while other more basic errors are flagged directly without being logged. I think this is fine if that is what you intended, but I wanted to be sure this is what you intended!

edaub (approved):
In addition to the other comments, a comment on the ``__init__`` and ``setup`` methods of ``AutoEmulate``: is the ``_init__`` method really doing anything useful? I can't imagine needing to create an instance of the class without wanting to affix my training data, so I'd prefer to set the data with ``__init__`` and then provide another interface like ``setup`` for modifying any of the fitting parameters after the fact.

mastoffel:
Yes, good question about `__init__()`. Initially, I tried to be close in style to existing packages, like [pycaret](https://pycaret.gitbook.io/docs/get-started/quickstart#oop-api) which has a very similar API. I'll have a close look at this and think about the rationale behind it.

autoemulate/hyperparam_searching.py
```diff
@@ -0,0 +1,175 @@
...
+    inbuilt_grid = get_model_params(model)
+    # check that all keys in param_grid are in the inbuilt grid
+    for key in param_grid.keys():
+        if key not in inbuilt_grid.keys():
+            raise ValueError(f"Invalid parameter: {key}")
+
+    return param_grid
+
+
+# if __name__ == "__main__":
```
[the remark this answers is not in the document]
mastoffel: Yes, deleted.

autoemulate/hyperparam_searching.py
mastoffel: Thanks, done. Allowing to pass a sklearn cv instance is already issue #116 .

autoemulate/hyperparam_searching.py
mastoffel: Good catch. Haven't carefully thought about logging yet as I'm a bit unsure what would be a good strategy here. But this is an issue too #36 .

PR #115, opened 2024-01-19
Add dimension reduction pre-processing step

### LINKED ISSUE #93: Add dimension reduction
For datasets with many inputs it would be good to implement dimension reduction (PCA, autoencoders) as an option.

mastoffel:
Adds arguments to reduce input data dimensionality. Should work with anything in [sklearn.decomposition](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.decomposition).

Resolves #93


edaub (approved):
I think this is fine. Only possible thing is what to do if dimension reduction fails for some reason (I don't know how likely this is)?

mastoffel:
> I think this is fine. Only possible thing is what to do if dimension reduction fails for some reason (I don't know how likely this is)?

Good point. Given that the input is checked thoroughly, hopefully not too much went wrong. I've tried a few things though, including extracting more PC's than input features and inputting a sparse matrix. Both gave informative error messages, so I think this is fine for now.

Issue #128, opened 2024-01-30
Optimise PyTorch MLP hyperparameter search


mastoffel:
MLP hyperparameter search is a little tricky as the space includes the architecture (layers etc.). We need to find a good hyperparameter grid subspace that works well with our datasets.

kallewesterling:
@bryanlimy @mastoffel Are there any updates on this issue?

bryanlimy:
I am playing with the hyperparameter grid and able to find a model that achieved >0.89 R**2 on `cardiac1`, but it seems like there is an issue passing `module__` arguments to `TorchModule`. I am debugging it now.

bryanlimy:
https://github.com/alan-turing-institute/autoemulate/pull/157

bryanlimy:
Completed in https://github.com/alan-turing-institute/autoemulate/pull/157

PR #157, opened 2024-02-12
`MLPModule` search space for `cardiac` datasets

bryanlimy:
- update `get_module` to return only the class, without initialization, so that subsequent calls to `set_params` can result `self.module`. 
  - this update fixed 2 test cases which we previously ignored
  - this also allows the `em.compare()` function to set the module hyperparameter properly
- use [`AdamW`](https://arxiv.org/abs/1711.05101) instead of `Adam` as default optimizer and set default learning rate to `1e-3` to align with PyTorch.
- updated `MLPModule` search space, add `batch_size`, `optimizer__weight_decay`, `module__hidden_activation`.

address Issue https://github.com/alan-turing-institute/autoemulate/issues/128

Example on `cardiac1` dataset
```
autoemulate - Performing grid search for NeuralNetTorch...
autoemulate - Best parameters for NeuralNetTorch: {'model__batch_size': 2, 'model__lr': 0.00031021307432157666, 'model__max_epochs': 70, 'model__module__hidden_activation': <class 'torch.nn.modules.activation.Tanh'>, 'model__module__hidden_sizes': (100, 100), 'model__optimizer__weight_decay': 0.04215478953678464}
autoemulate - Cross-validating NeuralNetTorch...
autoemulate - Parameters: {'module': <class 'autoemulate.emulators.neural_networks.mlp.MLPModule'>, 'criterion': <class 'torch.nn.modules.loss.MSELoss'>, 'optimizer': <class 'torch.optim.adamw.AdamW'>, 'lr': 0.00031021307432157666, 'max_epochs': 70, 'batch_size': 2, 'iterator_train': <class 'torch.utils.data.dataloader.DataLoader'>, 'iterator_valid': <class 'torch.utils.data.dataloader.DataLoader'>, 'dataset': <class 'skorch.dataset.Dataset'>, 'train_split': False, 'callbacks': [<autoemulate.emulators.neural_net_torch.InputShapeSetter object at 0x297371bd0>], 'predict_nonlinearity': 'auto', 'warm_start': False, 'verbose': 0, 'device': 'cpu', 'compile': False, 'use_caching': 'auto', '_params_to_validate': {'optimizer__weight_decay', 'module__input_size', 'module__hidden_sizes', 'module__output_size', 'iterator_train__shuffle', 'module__hidden_activation'}, 'module__input_size': 18, 'module__output_size': 4, 'optimizer__weight_decay': 0.0001, 'iterator_train__shuffle': True, 'module__hidden_activation': <class 'torch.nn.modules.activation.Tanh'>, 'module__hidden_sizes': (100, 100), 'callbacks__epoch_timer': <skorch.callbacks.logging.EpochTimer object at 0x29a787e10>, 'callbacks__train_loss': <skorch.callbacks.scoring.PassthroughScoring object at 0x29ae24ad0>, 'callbacks__train_loss__name': 'train_loss', 'callbacks__train_loss__lower_is_better': True, 'callbacks__train_loss__on_train': True, 'callbacks__valid_loss': <skorch.callbacks.scoring.PassthroughScoring object at 0x29b45dc50>, 'callbacks__valid_loss__name': 'valid_loss', 'callbacks__valid_loss__lower_is_better': True, 'callbacks__valid_loss__on_train': False, 'callbacks__InputShapeSetter': <autoemulate.emulators.neural_net_torch.InputShapeSetter object at 0x297371bd0>, 'callbacks__print_log': <skorch.callbacks.logging.PrintLog object at 0x29a7967d0>, 'callbacks__print_log__keys_ignored': None, 'callbacks__print_log__sink': <built-in function print>, 'callbacks__print_log__tablefmt': 'simple', 'callbacks__print_log__floatfmt': '.4f', 'callbacks__print_log__stralign': 'right'}
autoemulate - NeuralNetTorch is the best model with R^2 = 0.686
```

p.s. I notice the variation between runs (`em.compare()`) are quite big, are we setting a seed somewhere?

### DIFF
--- a/autoemulate/emulators/neural_networks/mlp.py
+++ b/autoemulate/emulators/neural_networks/mlp.py
@@ -49,6 +49,7 @@ class MLPModule(TorchModule):
                 nn.Sigmoid,
                 nn.GELU,
             ],
+            "optimizer": [torch.optim.AdamW, torch.optim.SGD],
             "optimizer__weight_decay": (1 / 10 ** np.arange(1, 9)).tolist(),
         }
         match search_type:
--- a/autoemulate/emulators/neural_networks/mlp.py
+++ b/autoemulate/emulators/neural_networks/mlp.py
@@ -18,7 +18,8 @@ class MLPModule(TorchModule):
         input_size: int = None,
         output_size: int = None,
         random_state: int = None,
-        hidden_sizes: Tuple[int] = (100,),
+        hidden_layers: int = 1,
+        hidden_size: int = 100,
         hidden_activation: Tuple[callable] = nn.ReLU,
     ):
         super(MLPModule, self).__init__(
@@ -28,7 +29,8 @@ class MLPModule(TorchModule):
             random_state=random_state,
         )
         modules = []
-        for hidden_size in hidden_sizes:
+        assert hidden_layers >= 1
+        for _ in range(hidden_layers):
             modules.append(nn.Linear(in_features=input_size, out_features=hidden_size))
             modules.append(hidden_activation())
             input_size = hidden_size
@@ -37,18 +39,23 @@ class MLPModule(TorchModule):
 
     def get_grid_params(self, search_type: str = "random"):
         param_space = {
-            "lr": loguniform(1e-06, 0.01),
             "max_epochs": np.arange(10, 110, 10).tolist(),
             "batch_size": np.arange(2, 128, 2).tolist(),
-            "module__hidden_sizes": [(50,), (100,), (100, 100), (100, 200), (200, 200)],
-            "module__hidden_activation": [nn.ReLU, nn.Tanh, nn.Sigmoid, nn.GELU],
-            "optimizer__weight_decay": (1 / 10 ** np.arange(1, 10)).tolist(),
+            "module__hidden_layers": np.arange(1, 4).tolist(),
+            "module__hidden_size": np.arange(50, 250, 50).tolist(),
+            "module__hidden_activation": [
+                nn.ReLU,
+                nn.Tanh,
+                nn.Sigmoid,
+                nn.GELU,
+            ],
+            "optimizer__weight_decay": (1 / 10 ** np.arange(1, 9)).tolist(),
         }
         match search_type:
             case "random":
-                pass
+                param_space |= {"lr": loguniform(1e-06, 1e-2)}
             case "bayes":
-                pass
+                param_space |= {"lr": Real(1e-06, 1e-2, prior="log-uniform")}
             case _:
                 raise ValueError(f"Invalid search type: {search_type}")
 
--- a/autoemulate/emulators/neural_networks/mlp.py
+++ b/autoemulate/emulators/neural_networks/mlp.py
@@ -19,6 +19,7 @@ class MLPModule(TorchModule):
         output_size: int = None,
         random_state: int = None,
         hidden_sizes: Tuple[int] = (100,),
+        hidden_activation: Tuple[callable] = nn.ReLU,
     ):
         super(MLPModule, self).__init__(
             module_name="mlp",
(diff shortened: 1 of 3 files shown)


autoemulate/emulators/neural_networks/mlp.py
```diff
@@ -28,30 +30,25 @@ def __init__(
...
 
     def get_grid_params(self, search_type: str = "random"):
+        param_space = {
+            "lr": loguniform(1e-06, 0.01),
+            "max_epochs": np.arange(10, 110, 10).tolist(),
+            "batch_size": np.arange(2, 128, 2).tolist(),
+            "module__hidden_sizes": [(50,), (100,), (100, 100), (100, 200), (200, 200)],
+            "module__hidden_activation": [nn.ReLU, nn.Tanh, nn.Sigmoid, nn.GELU],
+            "optimizer__weight_decay": (1 / 10 ** np.arange(1, 10)).tolist(),
+        }
```
mastoffel, line 46: This currently fails for the [bayes search](https://scikit-optimize.github.io/stable/modules/generated/skopt.BayesSearchCV.html) (see below), which needs a slightly different `param_space` definition (see other models). It's a bit annoying but we have essentially define the `search_space` twice for now.

```
em = AutoEmulate()
em.setup(X, y, param_search_type="bayes", model_subset=["NeuralNetTorch"], param_search=True)
em.compare()
```

mastoffel (commented):
Again, great work @bryanlimy. I'm a little tight on time so haven't done testing yet, but wanted to provide one comment below.

autoemulate/emulators/neural_networks/mlp.py
bryanlimy: fixed the issue. also separated `module__hidden_size` and `module__hidden_layers` so that we can use `bayes` for it as well.

mastoffel:
@bryanlimy just thinking about the variation in `em.compare()`. With `param_search=False`, the variation is quite small I think (though there is some). With `param_search=True`, the variation is large, because we default on `param_search_iter=20`, which randomly searches through 20 different parameter combinations. I don't know whether we would want to set a seed here, but I think it would make sense to give the option at least.

autoemulate/emulators/neural_networks/mlp.py
```diff
@@ -26,32 +29,33 @@ def __init__(
...
+            "module__hidden_layers": np.arange(1, 4).tolist(),
+            "module__hidden_size": np.arange(50, 250, 50).tolist(),
+            "module__hidden_activation": [
+                nn.ReLU,
+                nn.Tanh,
+                nn.Sigmoid,
+                nn.GELU,
+            ],
+            "optimizer__weight_decay": (1 / 10 ** np.arange(1, 9)).tolist(),
+        }
```
mastoffel, line 53: This looks good. One thing I wondered is whether we should search over a few optimizers, in particular `lbfgs` (see [here](https://pytorch.org/docs/stable/generated/torch.optim.LBFGS.html) and [scikit-learn-doc](https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPRegressor.html). When I tested a few smaller datasets with the scikit learn MLP, `lbfgs` tended to be the best optimiser. However, it can become prohibitive for larger datasets (which we won't encounter very much I think).

mastoffel (commented):
Nice, bayes search works well now! Just one comment on optimisers below, then happy for you to merge!

bryanlimy:
> @bryanlimy just thinking about the variation in `em.compare()`. With `param_search=False`, the variation is quite small I think (though there is some). With `param_search=True`, the variation is large, because we default on `param_search_iter=20`, which randomly searches through 20 different parameter combinations. I don't know whether we would want to set a seed here, but I think it would make sense to give the option at least.

Yeah, I am getting consistent results with `bayes` by not `random`. Perhaps we can set the seed for random search such that user can reproduce the exact set of experiments but this is a non-issue in the meantime.

autoemulate/emulators/neural_networks/mlp.py
bryanlimy: I can add `optimizer: [torch.optim.AdamW, torch.optim.SGD]`. `LBFGS` doesn't support weight decay, we would have to setup conditional search space otherwise.

mastoffel: Ok, sounds good. Maybe we can open an issue to add LBFGS eventually.

kallewesterling:
Good work, guys!

Issue #165, opened 2024-02-13
Add second PyTorch model


mastoffel:
* doing a quick search for model architectures that have been used in surrogate/emulator modelling
* implementing the model (CNN?)

bryanlimy:
https://github.com/alan-turing-institute/autoemulate/pull/180

PR #180, opened 2024-02-22
Add RBF module 

bryanlimy:
- added `RFBModule` from https://github.com/rssalessio/PytorchRBFLayer/tree/main
  - quick experiments on the hyperparameter search space.
- set `module__input_size` and `module__output_size` default values to `None`
  - this requires us to initialize `module` within `fit` and `partial_fit` by inferring the input and output dimensions from `X` and `y`. 
    -  we need to ignore`check_estimators_empty_data_messages` test case because it expect the error message to contain the required size of the input data, however, at that point of the code, we don't know what is the expected input dimension.
- create `tests/test_torch.py` for all our custom test cases for `TorchModule` and `NeuralNetTorch`
- set `get_grid_params` as `staticmethod` so that we can obtain the search space for the TorchModule without initializing the model.

address https://github.com/alan-turing-institute/autoemulate/issues/165 and https://github.com/alan-turing-institute/autoemulate/issues/160

### DIFF
--- a/autoemulate/emulators/neural_networks/rbf.py
+++ b/autoemulate/emulators/neural_networks/rbf.py
@@ -268,9 +268,20 @@ class RBFLayer(nn.Module):
 
 
 class RBFModule(TorchModule):
-    """
-    Radial Basis Function Layer module for NeuralNetRegressor
+    """Radial Basis Function Layer
 
+    Wrapper around the Radial Basis Function Layer package https://github.com/rssalessio/PytorchRBFLayer.
+
+    Parameters
+    ---
+    radial_function: callable
+        a radial basis function that returns a tensor of real values given a
+        tensor of real values. `rbf_gaussian`, `rbf_linear`, `rbf_multiquadric`,
+        `rbf_inverse_quadratic`, `rbf_inverse_multiquadric` are available.
+    norm_function: callable
+        a normalization function applied on the features. `l_norm` is available.
+    normalization: bool
+        if True applies the normalization trick to the RBF layer
     """
 
     def __init__(
--- a/autoemulate/emulators/neural_networks/rbf.py
+++ b/autoemulate/emulators/neural_networks/rbf.py
@@ -313,7 +313,8 @@ class RBFModule(TorchModule):
         ]
         self.model = nn.Sequential(*modules)
 
-    def get_grid_params(self, search_type: str = "random"):
+    @staticmethod
+    def get_grid_params(search_type: str = "random"):
         param_space = {
             "max_epochs": np.arange(10, 110, 10).tolist(),
             "batch_size": np.arange(2, 128, 2).tolist(),
--- a/autoemulate/emulators/neural_networks/rbf.py
+++ b/autoemulate/emulators/neural_networks/rbf.py
@@ -1,0 +1,331 @@
+from typing import Callable
+from typing import Tuple
+from typing import Union
+
+import numpy as np
+import torch
+from scipy.stats import loguniform
+from skopt.space import Integer
+from skopt.space import Real
+from torch import nn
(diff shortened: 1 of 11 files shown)


tests/test_torch.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+    nn_torch_model.fit(X, y)
+    assert nn_torch_model.module__input_size == input_size
+    assert nn_torch_model.n_features_in_ == input_size
+    assert nn_torch_model.module_.model[0].in_features == input_size
+    assert nn_torch_model.module__output_size == output_size
+    assert nn_torch_model.module_.model[-1].out_features == output_size
+
+
+def test_nn_torch_module_methods():
```
mastoffel, line 92: Nice

autoemulate/emulators/neural_networks/rbf.py
```diff
@@ -0,0 +1,331 @@
...
+    def get_shapes(self):
+        """Returns the shape parameters"""
+        return self.log_shapes.detach().exp()
+
+
+class RBFModule(TorchModule):
+    """
+    Radial Basis Function Layer module for NeuralNetRegressor
+
+    """
```
mastoffel, line 274: Would be good to have a description of the parameters here.

autoemulate/emulators/neural_networks/rbf.py
```diff
@@ -0,0 +1,331 @@
...
+            "module__radial_function": [
+                rbf_gaussian,
+                rbf_linear,
+                rbf_multiquadric,
+                rbf_inverse_quadratic,
+                rbf_inverse_multiquadric,
+            ],
+            "optimizer": [torch.optim.AdamW, torch.optim.SGD],
+            "optimizer__weight_decay": (1 / 10 ** np.arange(1, 9)).tolist(),
+        }
```
mastoffel, line 319: looks good, but doesn't work great yet. I don't have a good intuition whether radial basis function networks will work well at all for our datasets, but maybe we can try to optimise this in the next sprint.

tests/test_torch.py
mastoffel, line 1: Maybe a "high level" test for the UI would be good to add to see whether the NN's work well, like this:

```
def test_nn_torch_module_ui():
    input_size, output_size = 10, 2
    X = np.random.rand(100, input_size)
    y = np.random.rand(100, output_size)
    em = AutoEmulate()
    em.setup(X, y, model_subset=["NeuralNetTorch"])
    # check that compare does not raise an error
    best = em.compare()
    assert get_model_name(best) == "NeuralNetTorch"


def test_nn_torch_module_ui_param_search():
    input_size, output_size = 10, 2
    X = np.random.rand(100, input_size)
    y = np.random.rand(100, output_size)
    em = AutoEmulate()
    em.setup(X, y, model_subset=["NeuralNetTorch"], param_search=True)
    # check that compare does not raise an error
    best = em.compare()
    assert get_model_name(best) == "NeuralNetTorch"
```

mastoffel (commented):
OK, this is great @bryanlimy , and works well. I think this can be merged after the comments are addressed. 

What doesn't work yet is to have both `mlp` and `rbf` in the `MODEL_REGISTRY` under `autoemulate.emulators.__init__ `. This is because 1) currently the models aren't initialised in the registry and 2) The names of models in `autoemulate` are currently simply the class names (which we can get with `get_model_name()` from `autoemulate.utils`), but both `mlp` and `rbf` have the same class name `NeuralNetTorch`. So to include both in `compare()`, we need to fix naming, but I think this is another PR.

Also, if you want to merge or rebase, there are a few changes in `main` including one that fixes 1) above.

mastoffel (approved):

PR #195, opened 2024-02-27
Improve user interface

### LINKED ISSUE #159: improve user interface
* get a nice table of `setup()` variables when `setup()` is called
* get a nice table of `models` with abbreviation and full name

kallewesterling:
Fixes #159.

### DIFF
--- a/autoemulate/compare.py
+++ b/autoemulate/compare.py
@@ -125,7 +125,12 @@ class AutoEmulate:
         self.n_jobs = n_jobs
         self.logger = _configure_logging(log_to_file=log_to_file)
         self.is_set_up = True
+        self.dim_reducer = dim_reducer
+        self.reduce_dim = reduce_dim
+        self.folds = folds
         self.cv_results = {}
+
+        self.print_settings()
 
     def _check_input(self, X, y):
         """Checks and possibly converts the input data.
@@ -198,7 +203,12 @@ class AutoEmulate:
         self.scores_df = pd.DataFrame(
             columns=["model", "metric", "fold", "score"]
         ).astype(
-            {"model": "object", "metric": "object", "fold": "int64", "score": "float64"}
+            {
+                "model": "object",
+                "metric": "object",
+                "fold": "int64",
+                "score": "float64",
+            }
         )
 
         for i in range(len(self.models)):
@@ -368,6 +378,74 @@ class AutoEmulate:
 
         return serialiser._load_model(path)
 
+    def print_settings(self):
+        if not self.is_set_up:
+            raise RuntimeError("Must run setup() before print_settings()")
+            return
+
+        models = "\n- " + "\n- ".join(
+            [
+                x[1].__class__.__name__
+                for pipeline in self.models
+                for x in pipeline.steps
+                if x[0] == "model"
+            ]
+        )
+        metrics = "\n- " + "\n- ".join([metric.__name__ for metric in self.metrics])
+
+        settings = pd.DataFrame(
+            [
+                str(self.X.shape),
+                str(self.y.shape),
+                str(self.train_idxs.shape),
+                str(self.test_idxs.shape),
+                str(self.param_search),
+                str(self.search_type),
+                str(self.param_search_iters),
+                str(self.scale),
+                str(
+                    self.scaler.__class__.__name__ if self.scaler is not None else None
+                ),
+                str(self.reduce_dim),
+                str(
+                    self.dim_reducer.__class__.__name__
+                    if self.dim_reducer is not None
+                    else None
+                ),
+                str(self.cv.__class__.__name__ if self.cv is not None else "None"),
+                str(self.folds),
+                str(self.n_jobs),
+            ],
+            index=[
+                "Simulation input shape (X)",
+                "Simulation output shape (y)",
+                "Training dataset shape (train_idxs)",
+                "Test dataset shape (test_idxs)",
+                "Do hyperparameter search (param_search)",
+                "Type of hyperparameter search (search_type)",
+                "# sampled parameter settings (param_search_iters)",
+                "Scale data before fitting (scale)",
+                "Scaler (scaler)",
+                "Dimensionality reduction before fitting (reduce_dim)",
+                "Dimensionality reduction method (dim_reducer)",
+                "Cross-validation strategy (cv)",
+                "# folds (folds)",
+                "# parallel jobs (n_jobs)",
+            ],
+            columns=["Values"],
+        )
+
+        settings_str = settings.to_string(index=True, header=False)
+        width = len(settings_str.split("\n")[0])
+
+        print("AutoEmulate is set up with the following settings:")
+        print("-" * width)
+        print(settings_str)
+        print("-" * width)
+        print("Models:" + models)
+        print("-" * width)
+        print("Metrics:" + metrics)
+
     def print_results(self, model=None, sort_by="r2"):
         """Print cv results.
 
@@ -470,5 +548,10 @@ class AutoEmulate:
             Number of columns in the plot grid for multi-output. Default is 2.
         """
         _plot_model(
-            model, self.X[self.test_idxs], self.y[self.test_idxs], plot, n_cols, figsize
+            model,
+            self.X[self.test_idxs],
+            self.y[self.test_idxs],
+            plot,
+            n_cols,
+            figsize,
         )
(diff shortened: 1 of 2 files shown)


autoemulate/compare.py
```diff
@@ -368,6 +378,74 @@ def load_model(self, path=None):
...
+                ),
+                str(self.cv.__class__.__name__ if self.cv is not None else "None"),
+                str(self.folds),
+                str(self.n_jobs),
+            ],
+            index=[
+                "Simulation input shape (X)",
+                "Simulation output shape (y)",
+                "Training dataset shape (train_idxs)",
+                "Test dataset shape (test_idxs)",
```
mastoffel, line 423: Just thinking, so these are just the indices, I think we can just say:

```
"# training set samples"
"# test set samples"

```
And do [0] above to get an `int` instead of a `tuple`.

autoemulate/compare.py
```diff
@@ -368,6 +378,74 @@ def load_model(self, path=None):
...
+            ]
+        )
+        metrics = "\n- " + "\n- ".join([metric.__name__ for metric in self.metrics])
+
+        settings = pd.DataFrame(
+            [
+                str(self.X.shape),
+                str(self.y.shape),
+                str(self.train_idxs.shape),
+                str(self.test_idxs.shape),
```
mastoffel, line 401:
```
str(self.train_idxs.shape[0]),
str(self.test_idxs.shape[0]),
```

To get an int instead of tuple, see below.

autoemulate/compare.py
```diff
@@ -368,6 +378,74 @@ def load_model(self, path=None):
 
         return serialiser._load_model(path)
 
+    def print_settings(self):
```
mastoffel, line 381: This works/looks great.

mastoffel: One thought: should we rename this to `print_setup` or do you have a reason to call this `settings` ?

mastoffel (commented):
Great job @kallewesterling, this is exactly what I was thinking about.

Two high level things:

1) Could we put this into a separate module to keep `compare` clean?
2) What would be great eventually (though no need for this now if you're tight on time) is to check whether `autoemulate` is run in a notebook and if so display the table with `display(HTML(df.to_html()))` from `from IPython.display`, which looks a bit nicer.

autoemulate/compare.py
kallewesterling: Great point!

kallewesterling: Addresed in https://github.com/alan-turing-institute/autoemulate/pull/195/commits/75d5ae61506dd76c7dd2ad3157c3cdc05c4e19e9

autoemulate/compare.py
kallewesterling: Addresed in https://github.com/alan-turing-institute/autoemulate/pull/195/commits/75d5ae61506dd76c7dd2ad3157c3cdc05c4e19e9

autoemulate/compare.py
kallewesterling: Addresed in https://github.com/alan-turing-institute/autoemulate/pull/195/commits/75d5ae61506dd76c7dd2ad3157c3cdc05c4e19e9

kallewesterling:
> Two high level things:
> 
> 1. Could we put this into a separate module to keep compare clean?
> 2. What would be great eventually (though no need for this now if you're tight on time) is to check whether autoemulate is run in a notebook and if so display the table with display(HTML(df.to_html())) from from IPython.display, which looks a bit nicer.

1. Do you mean to move it to, say, `utils.py`? I was thinking, as it pertains to the class specifically, it makes sense to have it as a method of the `compare.AutoEmulate` class?
2. This sounds like a great idea. I haven't found a super way of checking whether the script is run in a notebook or not.. 

   I have seen two different versions of trying to resolve it (see this source for instance) -- any idea which ones is preferrable?

   1.
   ```py
   try:
       __IPYTHON__
       _in_ipython_session = True
   except NameError:
       _in_ipython_session = False
   ```
   
   2.
   ```py
   try:
       from IPython import get_ipython
       ip = get_ipython()
       if ip is None:
               # we have IPython installed but not running from IPython
               return False
           else:
               from IPython.core.interactiveshell import InteractiveShell
               format = InteractiveShell.instance().display_formatter.format
               if len(format(_checkhtml, include="text/html")[0]):
                   # TODO: need to check for qtconsole here!
                   return True
               else:
                   return False
       except:
           # We do not even have IPython installed
           return False
   ```

mastoffel:
> > Two high level things:
> > 
> > 1. Could we put this into a separate module to keep compare clean?
> > 2. What would be great eventually (though no need for this now if you're tight on time) is to check whether autoemulate is run in a notebook and if so display the table with display(HTML(df.to_html())) from from IPython.display, which looks a bit nicer.
> 
> 1. Do you mean to move it to, say, `utils.py`? I was thinking, as it pertains to the class specifically, it makes sense to have it as a method of the `compare.AutoEmulate` class?
> 2. This sounds like a great idea. I haven't found a super way of checking whether the script is run in a notebook or not..
>    I have seen two different versions of trying to resolve it (see this source for instance) -- any idea which ones is preferrable?
>    
>    
>    
>    ```python
>    try:
>        __IPYTHON__
>        _in_ipython_session = True
>    except NameError:
>        _in_ipython_session = False
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    
>    
>    
>    ```python
>    try:
>        from IPython import get_ipython
>        ip = get_ipython()
>        if ip is None:
>                # we have IPython installed but not running from IPython
>                return False
>            else:
>                from IPython.core.interactiveshell import InteractiveShell
>                format = InteractiveShell.instance().display_formatter.format
>                if len(format(_checkhtml, include="text/html")[0]):
>                    # TODO: need to check for qtconsole here!
>                    return True
>                else:
>                    return False
>        except:
>            # We do not even have IPython installed
>            return False
>    ```

Ah sorry, I wasn't very clear. I think it's good to have it as a `autoemulate.compare` method, but I'd still put the majority of code in a separate module (say pretty_printing.py or so), and then import `_print_setup` for the `autoemulate.compare` `print_setup` method. This is just so that `autoemulate.compare` doesn't become too big with all it's methods.

Good question with the checking for whether it's run in a notebook. I couldn't find a super clear answer either. I'll have a closer look tomorrow, but both versions seem to work.

kallewesterling:
Fixed the high-level (1) above in https://github.com/alan-turing-institute/autoemulate/pull/195/commits/6dab835c6bf949bfbe151adc305cdb10271be535

kallewesterling:
Fixed (2) in https://github.com/alan-turing-institute/autoemulate/pull/195/commits/daec70f5b742868c240c24424f3860c575101188

kallewesterling:
OK @mastoffel I think that's all the stuff addressed! Check out the branch now :)

autoemulate/printing.py
```diff
@@ -34,3 +45,84 @@ def _print_cv_results(models, scores_df, model=None, sort_by="r2"):
...
+            "Do hyperparameter search (param_search)",
+            "Type of hyperparameter search (search_type)",
+            "# sampled parameter settings (param_search_iters)",
+            "Scale data before fitting (scale)",
+            "Scaler (scaler)",
+            "Dimensionality reduction before fitting (reduce_dim)",
+            "Dimensionality reduction method (dim_reducer)",
+            "Cross-validation strategy (cv)",
+            "# folds (folds)",
+            "# parallel jobs (n_jobs)",
```
mastoffel, line 108: one after thought here: Maybe this would be most useful if the argument names (rather than attribute names) would be in brackets. Would you agree?

If so, the two things to change are:

```
"# training set samples (train_idxs)",
"# test set samples (test_idxs)",
```
to 
` "test set size (test_set_size)" `
(and the value could be 20% or 0.2)

and 
`"Cross-validation strategy (cv)",`
to
`"Cross-validation strategy (fold_strategy)",`

mastoffel (approved):
Very nice @kallewesterling , notebook printing looks great, and `printing.py` is a good place for this.

Just one more comment below.

autoemulate/printing.py
kallewesterling: Ah, that makes sense. I was thinking that these should refer to any properties that exist on the object (hence, `cv` instead of `fold_strategy` as `self.cv` has been set; same with `train_idxs` and `test_idxs` as those are also existing as properties, whereas `test_set_size` is not a property on the object)... But since we're just showing the setup settings, I suppose it makes sense!

kallewesterling: See: https://github.com/alan-turing-institute/autoemulate/pull/195/commits/c0504defae06b2b9b7ca633a7f6a0b6824dfa3e2

mastoffel: ok, thanks @kallewesterling ! Feel free to merge!

Issue #190, opened 2024-02-27
Including multiple PyTorch models in `compare()`


mastoffel:
In the MODEL_REGISTRY under `autoemulate.emulators.__init__py`, we are adding models like this:

```
MODEL_REGISTRY = {
 "GaussianProcessSk": GaussianProcessSk(),
 "SupportVectorMachines": SupportVectorMachines(),
"NeuralNetTorch": NeuralNetTorch(module="mlp")
}
```

A second neural network would be added like this:

```
MODEL_REGISTRY = {
 "GaussianProcessSk": GaussianProcessSk(),
 "SupportVectorMachines": SupportVectorMachines(),
"NeuralNetTorch": NeuralNetTorch(module="mlp")
"NeuralNetTorch": NeuralNetTorch(module="rbf")
}
```

Which currently fails. The reason is that there are a range of checks which use `get_model_name()`, a function that accesses the class name. So that function needs to be changed and different PyTorch networks renamed in the `MODEL_REGISTRY`. This means that a few methods/functions will likely need an additional input which is a dict mapping model names. See also #175

bryanlimy:
This is solved in https://github.com/alan-turing-institute/autoemulate/pull/196 ?

mastoffel:
solved now

PR #196, opened 2024-02-29
Model names

### LINKED ISSUE #175: sort out model names
All model names are simply the underlying class names at the moment.

* It would be better to have a dictionary with short names: full name for each model, so that users can select models with short names.  
* There should also be a method `get_models` or similar that prints short name, long name, source 
* This should also resolve the problem that currently different PyTorch networks have the same name (NeuralNetTorch), which is why they can't be loaded at the same time

mastoffel:
* changes `self.models` to `dict` with `model_names : model`
* changes `get_model_name()` to get model name from that dict
* having two pytorch models with the same class works now

resolves #175 and #190


autoemulate/emulators/__init__.py
```diff
@@ -12,13 +12,14 @@
...
     "SecondOrderPolynomial": SecondOrderPolynomial(),
-    "RBF": RBF(),
+    "RadialBasisFunctions": RBF(),
     "RandomForest": RandomForest(),
     "GradientBoosting": GradientBoosting(),
-    "GaussianProcessSk": GaussianProcessSk(),
+    "GaussianProcesses": GaussianProcessSk(),
     "SupportVectorMachines": SupportVectorMachines(),
     "XGBoost": XGBoost(),
+    "NeuralNet": NeuralNetTorch(module="mlp"),
```
bryanlimy, line 21: Should we have a prefix of `NN` followed by `<module name>` for all `NeuralNetTorch` model? 

e.g.
- `"NNMLP": NeuralNetTorch(module="mlp")`
- `"NNRBF": NeuralNetTorch(module="rbf")`

so that
- users can immediately tell they belongs to `NeuralNetTorch`
- they will be close to each other when we sort all the model names alphabetically for printing. 
- in `pytest`, we can test every model that has prefix `NN` instead of hard-coding their names in our test cases.

bryanlimy (approved):
LGTM. 

Have one suggestion for `NeuralNetTorch` model names.

Also, do you think it would be useful to have a getter function / attribute for all emulators that return the model name? e.g.
```
> model = NeuralNetTorch(module='mlp')
> model.model_name
NNMLP
> model = RandomForest()
> model.model_name 
RandomForest
```

autoemulate/emulators/__init__.py
kallewesterling: I like this suggestion - what do you think @mastoffel?

kallewesterling:
> LGTM.
> 
> Have one suggestion for `NeuralNetTorch` model names.
> 
> Also, do you think it would be useful to have a getter function / attribute for all emulators that return the model name? e.g.
> 
> ```
> > model = NeuralNetTorch(module='mlp')
> > model.model_name
> NNMLP
> > model = RandomForest()
> > model.model_name 
> RandomForest
> ```

This would be easy to implement through accessing `<Class>.__class__.__name__` in a simple implementation like this:

```py
@property
def model_name(self):
    return self.__class__.__name__
```

...at least I think...?

bryanlimy:
> > LGTM.
> > Have one suggestion for `NeuralNetTorch` model names.
> > Also, do you think it would be useful to have a getter function / attribute for all emulators that return the model name? e.g.
> > ```
> > > model = NeuralNetTorch(module='mlp')
> > > model.model_name
> > NNMLP
> > > model = RandomForest()
> > > model.model_name 
> > RandomForest
> > ```
> 
> This would be easy to implement through accessing `<Class>.__class__.__name__` in a simple implementation like this:
> 
> ```python
> @property
> def model_name(self):
>     return self.__class__.__name__
> ```
> 
> ...at least I think...?

`NeuralNetTorch(module='mlp')` and `NeuralNetTorch(module='rbf')` would have the same name in this case. We can have prefix + module like `model_name = NN + upper(self.module)`, which will lead to `NNMLP`, `NNRBF` etc.

autoemulate/emulators/__init__.py
mastoffel: both good suggestions! I'm on it.

mastoffel:
Ok, implemented both of @bryanlimy's suggestions. 
* Each model now has a `model_name` property method, and the `get_model_name` function from `utils`  now retrieves the model name even if the model is wrapped in a `Pipeline` and `Multioutput`. 
* Torch model names are now: ` def model_name(self):
        return f"NN{self.module_name.capitalize()}"` 
* did a few other renames to make things cleaner

mastoffel:
Maybe would be good if you could quickly have an eye on this too @bryanlimy ! Lots of changes here, but I think it's mostly fine.

bryanlimy:
LGTM!

Issue #127, opened 2024-01-29
Make logging work well for parallel processes


kallewesterling:

kallewesterling:
@mastoffel I need some more context here, would you be able to provide it? What's the current problem? How would we size this issue?

mastoffel:
actually, I think we should close this for now. Logging currently works in parallel, but likely to be revisited when we tackle #36

kallewesterling:
This is already marked as an issue that leads to the closing of #36 so I'd leave it open but take off `ai-uk` label perhaps?

Issue #126, opened 2024-01-29
Improve logging to file


kallewesterling:
We could allow writing only warnings, only non warnings etc. to different files.

kallewesterling:
Resolved with https://github.com/alan-turing-institute/autoemulate/pull/200

PR #200, opened 2024-03-06
Logging

### LINKED ISSUE #68: Log when models fail
Some models will inevitably fail for some datasets. For example, `SecondOrderPolynomial()` will fail when there's too few samples. Implement logging to catch those errors, and allow to be verbose to be able to see them. Arguably, the default should be hiding the actual error but displaying that the model failed, to keep it simple.

mastoffel:
Improves logging and adds a progress bar to `compare()`.

* `verbose` argument with 0-2 controls logging to a `StreamHandler` (what the user sees in the terminal)
* `log_to_file` now logs everything to file (not controlled by verbose, just everything)
* all warnings from model fitting / cv / param search are captured too
* `logger.exception` now logs errors alongside tracebacks
* `compare` now has a progress bar with `tqdm`, which works in notebooks and terminal

resolves #68 #126 #127


mastoffel:
@kallewesterling @bryanlimy if you two are busy, no worries, and I would just go ahead and merge this.

kallewesterling:
Yeah I think it looks pretty good. Just haven't had a chance to test it out myself but I trust that you have done so! Happy for you to merge.

mastoffel:
> LGTM!
> 
> Should we allow user to set where to log the file? Or at least make a print statement that logs are being stored to `autoemulate.log`.

Now allowing users to provide a path/filename for the log file, thanks @bryanlimy

Issue #222, opened 2024-08-01
CV is not done on the optimised hyper-parameters but on the default


MaxBalmus:
### Description

In the `compare` method of `autoemulate`, the results of hyperparameter optimisation are not fed into the CV step, rather the default value is used.

### Reproduction steps

```bash
1. When doing the model setup, set `log_to_file` to true.
2. Set `param_search` to true.
3. In autoemulate.log the `Best parameter for model` and `Parameters` (CV) are different.
```

### Version

0.1.0.post1  

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### OS

Linux

mastoffel:
fixed in #224

PR #224, opened 2024-08-09
Autoemulate.compare: proposed fix for issue #222

MaxBalmus:


autoemulate/compare.py
```diff
@@ -224,17 +224,27 @@ def compare(self):
...
-                            metrics=self.metrics,
-                            n_jobs=self.n_jobs,
-                            logger=self.logger,
-                        )
+                            # run cross validation with model that uses the optimized hyperparameters
+                            fitted_model, cv_results = _run_cv(
+                                X=self.X[self.train_idxs],
+                                y=self.y[self.train_idxs],
+                                cv=self.cross_validator,
+                                model=self.models[i],
```
mastoffel, line 232: yes, that's it

autoemulate/compare.py
```diff
@@ -224,17 +224,27 @@ def compare(self):
...
+                            # run cross validation
+                            fitted_model, cv_results = _run_cv(
+                                X=self.X[self.train_idxs],
+                                y=self.y[self.train_idxs],
+                                cv=self.cross_validator,
+                                model=model,
+                                metrics=self.metrics,
+                                n_jobs=self.n_jobs,
+                                logger=self.logger,
+                            )
```
mastoffel, line 247: I don't think we need this block, `self.models[i]` will be equal to `model` without parameter search and will contain the model with the updated hyperparameters with parameter search, so we just need the block above.

mastoffel (changes requested):
Thanks for spotting and fixing this @MaxBalmus !

Also, would you mind adding a test for this to `test_compare.py` ? I think something like this should work:
```
# --------------- test correct hyperparameter updating ------------------
def test_param_search_updates_models(ae, Xy):
    X, y = Xy
    ae.setup(X, y, model_subset=["RandomForest"], param_search=True,
             param_search_iters=5)
    params_before = ae.models[0].get_params() # just one model, so index with 0
    ae.compare()
    params_after = ae.models[0].get_params()
    assert params_before != params_after

def test_model_params_equal_wo_param_search(ae, Xy):
    X, y = Xy
    ae.setup(X, y, model_subset=["RandomForest"])
    params_before = ae.models[0].get_params()
    ae.compare()
    params_after = ae.models[0].get_params()
    assert params_before == params_after
```

tests can be run with `pytest`: `pytest tests/test_compare.py::test_param_search_updates_models`

autoemulate/compare.py
MaxBalmus: That's true. I made the change to mix the two versions so it does what we wanted.

MaxBalmus:
Silly question: how do I run the tests locally?

mastoffel:
> Silly question: how do I run the tests locally?

not a silly question! Run them with `pytest` in the terminal in the AutoEmulate directory:

`pytest tests/test_compare.py` -> run all tests in the file
`pytest tests/test_compare.py::specific_test` -> run specific_test in that file
`pytest -s tests/test_compare.py` -> show print statements in tests

let me know if something doesn't work!

mastoffel:
@all-contributors please add @MaxBalmus for code, bug

mastoffel (approved):
looks great!

PR #243, opened 2024-09-17
Expanded guassian_process_torch hyperparameter space

MaxBalmus:
1) Expanded the hyper parameter space of the torch GPEs to include more expensive, but also potentially more expensive cases:
    - added `ScaleKernel(MatternKernel())+ConstantKernel()`
    - set `ard_num_dims` to be the number of inputs for the kernels where the is possible (this allows the kernels to adjust the lengthscale based on individual parameters)
    - replaced `ConstantMean()` and `ZeroMean()` from the `mean_module` with the higher order `LinearMean()` and  `PolyMean(degree=2)` 
    - `PolyMean` is a new class of `mean_module` found in `autoemulatate.emulators.gaussian_process_utils`
    - `process_param_space`: added new parameter `input_dim` (default value 1) equal the number of dimensions of the input. This addition is necessary to be able to define the value of `ard_num_dims`.

2) Defined a new callback `EarlyStoppingMax` found in  `autoemulatate.emulators.gaussian_process_utils`:
    - the new class is derived `skorch.callbacks.EarlyStopping`
    - it fixes a bug which assumes that the cost function is always positive which can break the monotonicity of the cost function threshold 
    - a pull request was created on `skorch` to fix the original bug and is currently under review.


autoemulate/emulators/gaussian_processs_utils/polynomial_features.py
```diff
@@ -0,0 +1,42 @@
+import itertools
+
+import numpy as np
+import torch
+
+
+class PolynomialFeatures:
```
mastoffel, line 7: again, having a docstring and tests would be great here.

autoemulate/emulators/gaussian_process_torch.py
```diff
@@ -189,23 +209,90 @@ def predict(self, X, return_std=False):
...
+                    gpytorch.kernels.RBFKernel(ard_num_dims=input_dim).initialize(
+                        lengthscale=1.0
+                    ),
+                    gpytorch.kernels.MaternKernel(nu=2.5, ard_num_dims=input_dim),
+                    gpytorch.kernels.MaternKernel(nu=1.5, ard_num_dims=input_dim),
+                    gpytorch.kernels.PeriodicKernel(),
+                    gpytorch.kernels.RQKernel(ard_num_dims=input_dim),
+                    gpytorch.kernels.ScaleKernel(
+                        gpytorch.kernels.MaternKernel(ard_num_dims=input_dim)
+                    )
```
mastoffel, line 227: so the gp here is a multitask gp, and in the [docs](https://docs.gpytorch.ai/en/stable/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.html) they say 

> You may also notice that we don’t use a ScaleKernel, since the MultitaskKernel will do some scaling for us. (This way we’re not overparameterizing the kernel.)

So I think this can be taken out for now.  However, I'm just thinking about how to best have multiple gpytorch models, and then we'll have a gp with independent output dimensions where we'll wrap the kernels into ScaleKernels.

Also, on a side-note, I'd be curious about your experience with combined kernels (multiplicative/additive) and whether there's anything else we should add.

aranas:
>     4. it'd be great to add docstrings and tests where possible (see below for details)

Re testing, I personally would be very keen on some upskilling re best practices around code testing. I know there are many resources around code testing, see here for one developed at the Turing & Turing Way community: https://book.the-turing-way.org/reproducible-research/testing.

I wonder, @MaxBalmus  might this also be of interest to you and other researchers in TRIC? And @mastoffel would there be capacity to do a small session on this from REG side? apart from general upskilling for TRIC researchers re writing good tests, we could use this as an opportunity to develop a paragraph on contributor requirements re testing for autoemulate specifically. What do you think?

also to note, there is an upcoming [Turing Way book dash](https://docs.google.com/forms/d/e/1FAIpQLSemtUGORZFBdc6Cqjn4NpWKJd8qXI9qLvPEt_AtZJ7KTp-Bow/viewform) that could be an opportunity to run this session using (and potentially improving!) The Turing Way resource specifically.

mastoffel:
@aranas I'm sure there would be capacity if there's interest. Happy to chat about this. Lots of possibilities here, like an interactive session or just a talk etc. . I haven't read the Turing Way bits on testing yet, but will definitely do it now, thanks for the link!

autoemulate/emulators/gaussian_process_torch.py
MaxBalmus: I wasn't aware of that `MultitaskKernel` was doing some scaling.

I have a fairly limited understanding of the advantage of using combined kernels. From what I have seen there isn't any clear advantage, but I might be proven wrong.

autoemulate/emulators/gaussian_process_torch.py
```diff
@@ -109,6 +121,9 @@ def fit(self, X, y):
             self._y_train_std = _handle_zeros_in_scale(np.std(y, axis=0), copy=False)
             y = (y - self._y_train_mean) / self._y_train_std
 
+        X_train, X_val, y_train, y_val = train_test_split(X, y, test_size=0.2)
+        train_split = predefined_split(Dataset(X_val, y_val))
```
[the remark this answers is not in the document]
MaxBalmus: Yeah I see your point. Let's open a new PR.

autoemulate/emulators/gaussian_processs_utils/early_stopping_criterion.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
+from skorch.callbacks import EarlyStopping
+
+
+class EarlyStoppingMax(EarlyStopping):
```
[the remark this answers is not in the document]
MaxBalmus: Yeah that sounds reasonable. The skorch pull request is #1065: https://github.com/skorch-dev/skorch/pull/1065

autoemulate/emulators/gaussian_processs_utils/polynomial_features.py
MaxBalmus: Submitted the docstrings but to be honest I am not 100% how to go about making the tests.

mastoffel:
@MaxBalmus let me know if this is ready again to review!

PR #260, opened 2024-10-16
Sensitivity analysis

mastoffel:
Global sensitivity analysis using sobol indices, and plotting functions.


mastoffel:
@MaxBalmus @aranas just tagging you because of the tutorial you are writing. Feel free / let me know if you'd like to give comments / suggestions / review.  Otherwise I'd merge it tomorrow or Thursday. Here's how it will work:

This is how it currently works:

```
em = AutoEmulate()
em.setup(X, y)
em.compare()

emulator = em.get_model("gp")

# sensitivity analysis
problem = {
        'num_vars': 2,
        'names':  ['x1', 'x2]
        'bounds': [[0.1, 0.5]. [0.01, 0.2]]
}
Si = sensitivity_analysis(mod, problem,  as_df=True)
plot_sensitivity_analysis(Si, index="S1")
```

MaxBalmus:
At a glance it looks great. I will try to look more in depth.

Just a small feature request (sorry if is is a bit pedantic): since we already have `X` defined, and that is usually a data frame or derived from a dataframe, can we have a `problem_generator` function that read `X` and writes `problem` for you? The bounds could be derived from  the `min`/`max` of `X`.

mastoffel:
> At a glance it looks great. I will try to look more in depth.
> 
> Just a small feature request (sorry if is is a bit pedantic): since we already have `X` defined, and that is usually a data frame or derived from a dataframe, can we have a `problem_generator` function that read `X` and writes `problem` for you? The bounds could be derived from the `min`/`max` of `X`.

Yes, that sounds good!

PR #263, opened 2024-11-04
fix: correct link in comment

aranas:
just a minor fix as the linked address for projectile motion simulation has changed


mastoffel (approved):
Thanks @aranas

Issue #287, opened 2025-01-27
JOSS REVIEW: Community guidelines


willu47:
Part of [JOSS REVIEW](https://github.com/openjournals/joss-reviews/issues/7626#)

- [x] There are broken links in the `CONTRIBUTING.md` file.  Neither of the links in the [how to contribute](https://github.com/alan-turing-institute/autoemulate/blob/main/CONTRIBUTING.md#how-to-contribute) section work (these sections work in the documentation, I guess because the file is inserted into the relevant section).
- [x] There are no instructions on how to make a release or to deploy the package to PyPi (essential docs for maintainers)
- [x] Coupled with the above, I strongly encourage setting up a continuous deployment pipeline for deploying the package to PyPi pushing a tag (or making a release on Github), which greatly eases the process!

mastoffel:
The links are fixed, there is now a [release section](https://github.com/alan-turing-institute/autoemulate/blob/main/docs/community/release.md) in the docs, and we've added a [gh-actions workflow](https://github.com/alan-turing-institute/autoemulate/blob/main/.github/workflows/release.yaml) to automate the release upon pushing a tag, thanks for the suggestion @willu47!

PR #296, opened 2025-02-05
Community

mastoffel:
* fix links
* create gh-action to publish to pypi
* add release docs for maintainers

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,32 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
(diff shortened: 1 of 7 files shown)


mastoffel:
@marjanfamili @radka-j . I've added a gh-action workflow to automatically release the package on PyPi when pushing a new version tag as suggested in #287. To make this clear, I added a guide in the documentation (docs/community/release.md). I'm just wondering now whether this should just be under "community" as releasing requires maintainer rights, or whether we should have a seperate maintainer section or so for this. Would do you think?

marjanfamili (approved):
LGTM

radka-j (approved):
Looks great! The updated links work for me.

Only thought I had was whether there are any versioning guidelines we follow that we can link to in the release section? There might not be, this stuff is probably always more subjective than anything else.

I think keeping the release docs in the community section as is or adding a new maintainer section are both equally good options so pick whichever you're most comfortable with.

mastoffel:
Thanks both! @radka-j I added [this guide](https://py-pkgs.org/07-releasing-versioning.html), which I think is simple and intuitive.

Issue #294, opened 2025-02-03
Add history matching


radka-j:
This is a useful baseline for model calibration. 

Eventually it would be great to have an example of an active learning loop:
- use history matching to identify "not yet ruled out" parameter regions 
- run simulator with parameter values from these regions
- retrain emulator with this new data
Repeat to further narrow down the likely parameter space...

radka-j:
Closed in #294

PR #297, opened 2025-02-12
added history matching and test

marjanfamili:
This PR adds 

1. history matching 
2. relevant tests
3. adds an example to 01_start tutorial 

#294


autoemulate/history_matching.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+
+
+def history_matching(obs, expectations, threshold=3.0, discrepancy=0.0, rank=1):
+    """
+    Perform history matching to compute implausibility and identify NROY and RO points.
+
+    Parameters:
+        obs (tuple): Observations as (mean, variance).
+        expectations (tuple): Predicted (mean, variance).
```
mastoffel, line 10: i find obs and expectations to be a bit confusing as terminology. In a way my expectations are really the obs (e.g. what I see in the literature), and the other thing are my model predictions. What do you think? should we rename those?

autoemulate/history_matching.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+
+
+def history_matching(obs, expectations, threshold=3.0, discrepancy=0.0, rank=1):
+    """
+    Perform history matching to compute implausibility and identify NROY and RO points.
```
mastoffel, line 6: is this the implementation from [mogp-emulator](https://mogp-emulator.readthedocs.io/en/latest/)? It also might be good to explain this a bit more in the docs, e.g. to mention that this does not work in waves, but is a single history matching run.

mastoffel (changes requested):
Very nice PR @marjanfamili,  great that you also added this to the tutorial already.

I've got a few suggestions below. Also, I saw a few typos in the tutorial text (can't mark them here as the notebook has a too larg diff).

.gitignore
```diff
@@ -2,6 +2,7 @@ notebooks/
 __pycache__/
 .pytest_cache/
 dist/
+my_tests/
```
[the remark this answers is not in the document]
marjanfamili: Oh, good point I will do that, thanks

autoemulate/history_matching.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+
+
+def history_matching(obs, expectations, threshold=3.0, discrepancy=0.0, rank=1):
```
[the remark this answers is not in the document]
marjanfamili: Length and type are clear for me how to test, but range of obs values and expectations Im not so sure about, how would I go about this to have something universal ?

autoemulate/history_matching.py
marjanfamili: no it is not from anywhere specific, just evaluating the general formula. but I could potentially test this history matching with their simulation and demo to see if I get a similar result ?

marjanfamili: I suppose I have only explained this in the Tutorial Notebook. This is the only place I see the functionalities of the code explained. I can add more description to the function documentation itself.

autoemulate/history_matching.py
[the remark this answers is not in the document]
mastoffel: Sorry, that wasn't clear. No range needed for `obs` or `expectations` only for the other parameters, like
`threshold` and `discrepancy` need only to be positive, and `rank` also needs to be an `int` , like so
`assert isinstance(rank, int), 
  assert rank > 0
`

autoemulate/history_matching.py
mastoffel: > no it is not from anywhere specific, just evaluating the general formula. but I could potentially test this history matching with their simulation and demo to see if I get a similar result ?

that would be great.

mastoffel: > I suppose I have only explained this in the Tutorial Notebook. This is the only place I see the functionalities of the code explained. I can add more description to the function documentation itself.

yes, adding a bit more explanation to the docstring would be great. I feel that history matching can mean a few different things so good to be clear when someone looks it up.

mastoffel:
sorry @marjanfamili, one thing missing: would you mind also adding this to the  [API reference](https://alan-turing-institute.github.io/autoemulate/reference/index.html), see [here](https://alan-turing-institute.github.io/autoemulate/community/contributing-docs.html) on how to do this. Basically add an `history_matching.rst` to `docs/reference`, then add to the `docs/_toc.yml` and then run `jupyter-book build docs --all` to see whether it works.

autoemulate/history_matching.py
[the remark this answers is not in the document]
marjanfamili: cool, I will add that in

Issue #291, opened 2025-02-03
Add "fit only pytorch models" flag


radka-j:
Allow users to easily select to only fit emulators that have a PyTorch backend (currently this is GPs and CNPs). This is useful in cases where a downstream task relies on this.

mastoffel:
Just adding here, that when enabling this function, we should deactivate data pre-processing as this is done through sci-kit learn `pipelines`. We can then throw a warning to tell the user to do this manually beforehand.

mastoffel:
A bit of a brain dump here so that we can discuss this: I've just had a look at this issue and #295 (i.e. running and extracting PyTorch models). It's not as straightforward as I thought. This is because both Neural Processes and GPs need objects/data outside of the PyTorch object itself and both are quite specific. `AutoEmulate` handles these things in the background, but by strapping away the estimator object and returning the pure `PyTorch` object, we lose this functionality and leave it to the user to figure it out.

1. The [CNP](https://github.com/alan-turing-institute/autoemulate/blob/main/autoemulate/emulators/neural_networks/cnp_module.py) forward method needs context and targets points:

`   def forward(self, X_context, y_context, X_target=None, context_mask=None):`

`AutoEmulate` internally just takes the training data as context points, so in the `predict` method the user only has to provide targets `X`. To do training, the `CNP` uses a [dataset](https://github.com/alan-turing-institute/autoemulate/blob/main/autoemulate/emulators/neural_networks/datasets.py) which coordinates the sampling and is slightly weird, as it creates a meta-dataset from a normal dataset. So to do further training, the user would need that object too I guess. For a `Attentive CNP` the user would also need `context_mask` for training (which the `PyTorch dataset` in `AutoEmulate` also takes care of. 

2. The [GP](https://github.com/alan-turing-institute/autoemulate/blob/main/autoemulate/emulators/neural_networks/gp_module.py)  PyTorch object returns a `MultivariateNormal` object, rather then posterior predictions. To get those, we need a `likelihood` function, see [here](https://docs.gpytorch.ai/en/stable/examples/01_Exact_GPs/Simple_GP_Regression.html). We can extract this one from the object though.

So the question is how to go ahead. A few thoughts:
- extract all relevant objects and return them in a `tuple`. This means all `PyTorch` models will have different outputs and might be confusing.
- extract only the main `PyTorch` object and write tutorials on how to use them further
- something else?

Would be great to get your input here @marjanfamili @radka-j

[...]

PR #298, opened 2025-02-12
PyTorch model extractor

mastoffel:
- adds a function to extract PyTorch models from `AutoEmulate` emulators
- emulators can be `pipelines`, `MultiOutputRegressors` etc., this function checks all the options and extracts the underlying PyTorch model where possible and throws an error for other models
- it also gives a message saying that datapreprocessing is better turned off when doing this and has to be done manually (as it can't be attached to the `PyTorch` model like it can be to a `sci-kit learn` model using a `pipeline`
- it returns the model in `eval` mode

- does _not_ yet include other objects as discussed in #291 . Maybe we leave that to the next PR?

### DIFF
--- a/autoemulate/emulators/conditional_neural_process.py
+++ b/autoemulate/emulators/conditional_neural_process.py
@@ -213,6 +213,7 @@ class ConditionalNeuralProcess(RegressorMixin, BaseEstimator):
             verbose=0,
         )
         self.model_.fit(X, y)
+        self.is_fitted_ = True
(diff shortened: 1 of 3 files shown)


marjanfamili:
Thanks a lot for this PR Martin. It looks great. Would this functionality be necessary if we change everything to PyTorch? is this a temporary measure until we do ?

marjanfamili (approved):
Looks great, thanks

PR #299, opened 2025-02-13
Mle

marjanfamili:
Screenshot 2025-02-13 at 15 22 44

### DIFF
--- a/autoemulate/mle.py
+++ b/autoemulate/mle.py
@@ -1,0 +1,94 @@
+import numpy as np
+import torch
+
+def negative_log_likelihood(Sigma, obs_mean, model_mean):
+    """
+    Compute the negative log-likelihood for a multivariate Gaussian using NumPy.
+
+    Args:
+        Sigma (numpy.ndarray): Covariance matrix (\(\Sigma(\theta)\)).
+        obs_mean (numpy.ndarray): Observed mean values (\(y\)).
+        model_mean (numpy.ndarray): Predicted mean values (\(m(\theta)\)).
+
+    Returns:
+        float: Negative log-likelihood value.
+    """
+    # Add numerical stability to Sigma
+    noise_term = 1e-5 * np.eye(Sigma.shape[0])
+    Sigma = Sigma + noise_term
+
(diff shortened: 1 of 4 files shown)


marjanfamili:
One issue with this implementation is how the kernel is calculated. currently the kernel is calculated between the output means from the GP. 

Could this be done more efficiently by extracting the Kernel from the PyTorch model ? 
Are all our models going to be PyTorch ?

autoemulate/mle.py
```diff
@@ -0,0 +1,96 @@
...
+    # Quadratic term: (y - m(θ))^T Σ(θ)^(-1) (y - m(θ))
+    quad_term = np.dot(diff.T, np.dot(Sigma_inv, diff))
+
+    # Negative log-likelihood
+    nll = 0.5 * (quad_term + log_det_Sigma + len(diff) * np.log(2 * np.pi))
+
+    return nll.item()
+
+
+def max_likelihood(parameters, model, obs, lr=0.01, epochs=1000, kernel_name="RBF"):
```
cisprague, line 38: Just to clarify my understanding, is the following correct?

**Inputs and Outputs:**  
  - `parameters` are the candidate inputs to the simulation model.
  - The simulation outputs are summarized by their mean (`obs_mean`) and variance (`obs_var`). For a deterministic simulation, you’d typically have only `obs_mean`.

**Emulator Predictions:**  
  - The emulator (or surrogate model) takes the `parameters` as inputs and returns a predicted mean (`pred_mean`).
  - Instead of relying on a separately predicted variance (`pred_var`), the kernel is used within the negative log-likelihood (NLL) computation to generate the corresponding covariance.

**NLL Computation and Optimization:**  
  - The NLL is computed for each candidate by comparing the observed outputs (`obs_mean`) against the predictive density defined by `pred_mean` and the kernel-derived covariance.
  - The candidate input with the lowest NLL (i.e., the one that is most plausible under the emulator’s predictive distribution) is selected.
  - That selected candidate is then further refined using gradient descent to minimize the NLL, effectively fine-tuning the input so that the emulator's output distribution aligns closely with the observation.

PR #310, opened 2025-02-21
Show plots

mastoffel:
- this is a PR in response to a JOSS comment in the [review thread](https://github.com/openjournals/joss-reviews/issues/7626)
- plots now always show, whether run from a notebook or a terminal (previously one had to use `plt.show` in the terminal)
- plotting functions now use `plt.show` and return the `fig` object when run from a terminal, but don't use `plt.show` when run from a notebook, as that would cause double plotting
- added a function to determine the plot layout to avoid empty space when plotting multiple plots 
- all tests needed monkeypatching to avoid having to manually close plotting windows while testing

### DIFF
--- a/autoemulate/plotting.py
+++ b/autoemulate/plotting.py
@@ -474,9 +474,7 @@ def _plot_model(
         ax.set_visible(False)
     plt.tight_layout()
 
-    # prevent double plotting in notebooks
-    plt.close(fig)
-    return fig
+    return _display_figure(fig)
 
 
 def _plot_Xy(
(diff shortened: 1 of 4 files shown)


autoemulate/plotting.py
```diff
@@ -61,6 +61,30 @@ def _predict_with_optional_std(model, X_test):
...
+        Number of plots to display
+    n_cols : int, optional
+        Maximum number of columns allowed, default is 3
+
+    Returns
+    -------
+    tuple
+        (n_rows, n_cols) for the subplot layout
+    """
+    if n_plots <= 0:
```
radka-j, line 79:
```suggestion
    if n_plots <= 1:
```

radka-j (approved):
LGTM! I've run the quick start example from the terminal and the plots showed as expected and the intro notebook still displays correctly as well.

Issue #327, opened 2025-03-07
Refactor: probabilistic outputs and uncertainty


sgreenbury:
Emulation can be defined as supporting uncertainty in outputs in contrast to surrogates that are deterministic. Uncertainty/probabilistic outputs to be considered primary during refactor.

Some potential options for including uncertainty in outputs:
- Implement distribution types (e.g. [GPyTorch's distribution](https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/distributions/distribution.py)), with two sub-options:
  - Implement a new distribution type
  - Import one from a package (e.g. GPyTorch above)
- Return mean (and stddev), e.g. sklearn's GP [predict](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor.predict)
- ...

sgreenbury:
- Relating to #345 - check if MultivariateNormal distinguishes diagonal and non-diag cases differently

sgreenbury:
> * Relating to [Refactor: integrating active learning and initial refactor #345](https://github.com/alan-turing-institute/autoemulate/issues/345) - check if MultivariateNormal distinguishes diagonal and non-diag cases differently

This relates to #430.

sgreenbury:
Closing this following initial types #341 and #411 and superseded by #436

Issue #322, opened 2025-03-06
Refactor: design base classes for common functionality


sgreenbury:
Having (abstract) base classes to focus commonality around (i.e. code to the interface) will make it easier to navigate the code base, document and implement functionality in parallel. Implementing type hints is related here too (https://github.com/alan-turing-institute/autoemulate/pull/187).

For example, for emulators we have initially discussed designing a common base class for all emulators and then subclasses specific to PyTorch and non-PyTorch emulators.

sgreenbury:
Aiming to capture initial ideas and discussions on this here with an API, exploratory code and next steps.

### Design aims
Considering the design principles (see #314) the following seem useful features to include here:
- Common API for emulation tasks for all models:
    - fit
    - predict
    - tune
    - cross-validation
    - others?
- "torch first" models with `nn.Module` and have a similar/familiar API compared with torch
- Enable a low-code mode (relates to the common API point above)
- Integration with:
    - Simulator in the loop #324
    - ActiveLearner #301
- Handling of the following model types:
    - Deterministic outputs (e.g. surrogates)
    - Probabilistic outpits (e.g. emulators)
    - Non-torch models (sklearn)
- Runtime composability: to be able to incorporate preprocessing (e.g. into pipelines), model construction at runtime seems like a useful feature

### Possible base classes and structure

#### Description and code
The below python code has the following features:
- `BaseEmulator` ABC: all emulators would implement this an means that downstream integration (such as model comparison class and active learner) can code to this interface.
- `BaseEmulator` can aim to support pytorch types directly in the interface ("torch-first") in contrast to the current sklearn-first API.
- `PyTorchBackend`: this subclasses `nn.Module` and the `BaseEmulator` (providing the backend)
- `PyTorchModel`: this subclasses the `PyTorchBackend` so given it subclasses the mixin backend and just needs to implement `.forward()`.

Below is an example of this with example implementations (the function and signatures just initial ideas, would be great to hear thoughts on this)

```python
from abc import ABC, abstractmethod
from torch import nn

class BaseEmulator(ABC):
	@abstractmethod
	def fit(self, dataloader: DataLoader, config: FitConfig):
		pass
	
	@abstractmethod
	def predict(self, dataloader: DataLoader, config: PredictConfig):
		pass

	@abstractmethod
	def tune(self, dataset: Dataset, config: TuneConfig):
		pass

	@abstractmethod
	def cross_validate(self, dataset: Dataset, config: CVConfig):
[...]

Issue #301, opened 2025-02-17
Add modular active learning


cisprague:
**Main idea**
1. Implement an active learning loop where input points are selected to query the simulation model, retrieve output points, and iteratively improve the emulator model(s). A sampling budget should also be taken into account.
2. Modularity: The setup should be agnostic to the specific active learning approach, ensuring compatibility with history matching (#294) and Bayesian methods, while also supporting deterministic emulator models (e.g., using dropout for uncertainty estimation, or heuristic policies).
3. Since each learner (or "agent") could have multiple emulators, we need to ensure that policy is designed to handle multi-emulator scenarios. It would also be nice to leave open the possiblity of the policy being learnable.

**Possible API (psuedocode)**
- Here's a proposed API for an active learning loop (with 1 emulator) that aims to support multiple methods. If any existing components should be integrated, let me know. Would love to hear thoughts.
- Multiple emulators could possibly be accomodated by [EnsembleModel](https://github.com/pytorch/botorch/blob/052d0d8046c101504a01cac28f3f56e8fa0c6fda/botorch/models/ensemble.py#L24), and deterministic models (e.g. dropouot) by [EnsemblePosterior](https://github.com/pytorch/botorch/blob/052d0d8046c101504a01cac28f3f56e8fa0c6fda/botorch/posteriors/ensemble.py#L18), also see [dropout NN posterior](https://github.com/pytorch/botorch/discussions/2127#discussion-5880474).

```python
class Learner:

    def __init__(self, emulator, simulator, input_space, max_obs):
        self.emulator = emulator # Surrogate model
        self.simulator = simulator # Callable wrapper around user's sim
        self.obs = [] # Buffer of input-output pairs
        self.max_obs = max_obs # Max number of observations to store

    def update(self, new_obs):
        # Could have different update logic here
        if len(self.obs) < self.max_obs:
            self.obs.append(new_obs)
        else:
            idx = random.randint(0, self.max_obs - 1)
            self.obs[idx] = new_obs

    def policy(self):
        # Method to select new query points, which will be
        # different for different strategies.
        # Should be adapted to multiple emulators
        raise NotImplementedError # or ABC

    def iterate(self, simulation_func):
        # One iteration of active learning
        self.emulator.fit()
        new_point = self.policy()
        new_output = self.simulator(new_point)
        self.update((new_point, new_output))

class Bayesian_Optimizer(Learner):

[...]

PR #335, opened 2025-03-10
316 create working active learning example with existing codebase

cisprague:
Added an active learning example --- just stream-based active learning for now.


sgreenbury:
This is looking great @cisprague - I really like the class structure and experimental set-up exploring the different algorithms! It is also very helpful to see this with the current code base in relation to #322 for the refactor.

A few comments from a first look in case useful:
- This might be very subjective but the term "query" for the method `.query()` returning a `bool` I found slightly unintuitive initially. I think this is since it could be interpreted as a noun i.e. the query to be run (as opposed to the verb whether to run the query on the simulator).
- It might be worth defining query rate precisely (i.e. proportion of queries that that will be run?)
- I like the introduction to the learning algorithms above the subclassing - it might be worth adding a short doc comment to each class describing the algorithm and the parameters to remind the reader as they move through the tutorial of how they relate to each other.
- For the metrics plotted - I wonder if it's worth adding a little more description for what each metric tells us about a learner.  For example: "Uncertainty in the emulator measured with the standard deviation where a greater rate of reduction is indicative of better learner performance since fewer queries are required."
- Related to this - I wonder if the $R^2$ would be worth plotting as metric too?

radka-j:
This looks great! 

I mostly had clarifying questions that @sgreenbury already outlined, it would be useful to just have a bit more detail in the notebook text as well as the docstrings. 

An additional one was that I am unsure about is what membership-based AL strategy is. I know it's not the main focus here but as long as we mention it, is it possible to expand/reword this slightly to make it clearer?

Personally I don't mind the term `query`, especially if it's well established in the active learning domain?

radka-j:
I think we should move all the active learning classes to the codebase and have the notebook import everything from autoemulate and act primarily as a tutorial for active learning, only describing the class structure as required. I guess for now the notebook might need to keep the toy Emulator/Simulator classes. Once we have an Emulator class from the refactor, we can make any small updates to how the active learners use it and the notebook as required.

radka-j (commented):
This is really great!!! I learned a lot 😊 

I only had a question about how the rolling window in the r2 computation is implemented. Also, as mentioned above, I'd be in favour of moving everything into the codebase in this PR but am happy if others think that should be handled separately. 

The only last thing is that there are now 2 notebooks here, we probably want to delete the old one?

sgreenbury:
These updates look really great @cisprague! I found the additional technical detail/diagrams all very helpful. I think moving the classes into the library as @radka-j suggests sounds great and will be helpful for revising in the refactor (related to #322, #341, #327). I like the idea for the automatic registry too!

For the $R^2$ plot, the window as a parameter discussed above sounds like it will help so that each learner is averaged on the same number of values making their series more comparable. For the difference in length of the series, perhaps adding another field to the `Stream` to capture the iteration number that each query occurred could be useful (i.e. `self.queried_iteration.append(len(self.queries))`)? This could then be used to place the $R^2$ values at the iteration associated with e.g. the right hand side of the window and the series could extend across all iterations.

Another option might be to optionally add validation/test data into the `Learner` or `Stream` API. Like validation data in [Keras](https://keras.io/api/models/model_training_apis/#fit-method)/[Lightning](https://lightning.ai/docs/pytorch/stable/common/trainer.html#fit). However, this seems beyond the scope of the `Learner` (since it is monitoring performance on additional held-out data) so seems like a less good option here.

Two other small comments:
- I think the following deps (`seaborn` and `anytree`) need to be added to the dev deps to run the notebook?
- We might consider replacing the `List` and `Tuple` type hints with `list` and `tuple` (see [ruff pep585](https://docs.astral.sh/ruff/rules/non-pep585-annotation/)) since we're targeting Python >= 3.10? I can open an issue to consider if additional lints would be helpful to have in the pre-commit / CI.

cisprague:
@radka-j and @sgreenbury, thank you for the comments. The updates in this PR were written against the previous version of the code. The updated tutorial notebook should be put there since the parent PR (#301) refactors the API. Closing the PR.

Issue #371, opened 2025-03-27
Refactor: update compare method


sgreenbury:
Following implementing the #336 (TODO: add link to comment), the compare method will specifically need to be refactored.

sgreenbury:
Might be helpful to have `self.train` in `Autoemulate` as `self.train_val` to distinguish from the `Tuner`'s train

sgreenbury:
Adding some notes from discussion with @radka-j @EdwinB12:
- Refit on whole dataset at end of loop for best model
- Handle class init for different models
- Add tqdm for loop
- Reconsider error handling

PR #382, opened 2025-04-01
Extend InputTypeMixin

radka-j:
The motivation for this PR is primarily to set us up for #371.

Specifically, this PR:
- moves the `InputTypeMixin` class from `experimental.emulators.base.py` to `experimental.data.py`
- renames `_convert()` to `_convert_to_dataloader()`
- adds `_convert_to_dataset()`
- adds `_random_split()`
- extends the `Tuner` to subclass `InputTypeMixin` 

It also does a bit of docstrings clean up in `experimental/` to make it consistent with the rest of the repository.

Questions:
- I am still a bit unsure where the input data should be kept as a dataset vs. loader. I suspect some of this will become clearer as we also implement `Autoemulate.compare()`.


autoemulate/experimental/data.py
```diff
@@ -0,0 +1,86 @@
...
+        elif isinstance(x, DataLoader) and y is not None:
+            raise ValueError(
+                f"Since x is already a DataLoader, expect y to be None, not {type(y)}."
+            )
+        else:
+            dataset = self._convert_to_dataset(x, y)
+            dataloader = DataLoader(dataset, batch_size=batch_size, shuffle=shuffle)
+        return dataloader
+
+    def _random_split(
```
radka-j, line 59: After talking to @sgreenbury, we agreed that `_random_split` should return a `Dataset` not `DataLoader`

autoemulate/experimental/data.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Parameters
+        ----------
+        dataset: Dataset
+            The data to split.
+        batch_size: int | None
+            The DataLoader batch_size. If None, sets batch_size to lenth of training
+            data. Defaults to None.
+        """
+        train, test = tuple(random_split(dataset, [0.8, 0.2]))
```
sgreenbury, line 73: Perhaps we can have the train size as a parameter in method

autoemulate/experimental/data.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+
+        return dataset
+
+    def _convert_to_dataloader(
+        self,
+        x: InputLike,
+        y: OutputLike | None = None,
+        batch_size: int = 16,
+        shuffle: bool = True,
```
radka-j, line 43: We should also make sure that the higher level classes that call this method include a shuffle argument to pass to it.

autoemulate/experimental/data.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _convert_to_dataset(
+        self,
+        x: InputLike,
+        y: OutputLike | None = None,
+    ) -> Dataset:
+        """
+        Convert input data to pytorch Dataset.
+        """
+        # Convert input to DataLoader if not already
```
radka-j, line 21:
```suggestion
        # Convert input to Dataset if not already
```

autoemulate/experimental/data.py
sgreenbury: As discussed, can leave as `tuple[DataLoader, DataLoader]` here but to be refactored in #371

autoemulate/experimental/tuner.py
```diff
@@ -24,20 +31,18 @@ class Tuner:
         The validation scores and parameter values used in each search iteration.
     """
 
-    def __init__(self, dataset: Dataset, n_iter: int):
+    def __init__(self, x: InputLike, y: OutputLike | None, n_iter: int):
```
EdwinB12, line 34: Should y be InputLike here insteadof OutputLike?

EdwinB12, line 34:
```suggestion
    def __init__(self, x: InputLike, y: InputLike | None, n_iter: int):
```

EdwinB12:
Sorry I was a bit late to this but some nice changes in here! Nice work!

autoemulate/experimental/tuner.py
sgreenbury: Thanks - yes agreed!

PR #383, opened 2025-04-02
Add typing

### LINKED ISSUE #341: Refactor: add typing
A further design principle: include type hints with implementation.

The following will be useful:
- [ ] Add to design principles
- [x] ~~Add/revisit type hints to current code base~~ Aim to add types to refactor code in first instance
- [x] Add type checker to the CI/pre-commit

sgreenbury:
Closes #341.

This PR:
- Adds pyright to pre-commit config for experimental subpaths only
- Fixes type lints


sgreenbury:
pre-commit passes locally but Looks like there is an issue with pyright in the pre-commit for the CI - I'll take a look at fixing this.

autoemulate/experimental/tuner.py
```diff
@@ -57,15 +59,31 @@ def run(self, model_class: type[Emulator]) -> tuple[list[float], list[ModelConfi
...
 
             # evaluate
             y_pred = m.predict(val_x)
-            score = self.score_f(val_y, y_pred.detach().numpy())
+            # handle types
+            if isinstance(y_pred, TensorLike):
+                score = self.score_f(val_y, y_pred.detach().numpy())
+            elif isinstance(y_pred, DistributionLike):
+                score = self.score_f(val_y, y_pred.mean.detach().numpy())
+            elif (
```
radka-j, line 77: when is a tuple returned?

autoemulate/experimental/tuner.py
```diff
@@ -57,15 +59,31 @@ def run(self, model_class: type[Emulator]) -> tuple[list[float], list[ModelConfi
...
                 m = model_class(**params_sample)
+                assert isinstance(m, Emulator)
                 # TODO: check if pass as dataloader
                 m.fit(train_x, train_y)
 
             # evaluate
             y_pred = m.predict(val_x)
-            score = self.score_f(val_y, y_pred.detach().numpy())
+            # handle types
+            if isinstance(y_pred, TensorLike):
```
radka-j, line 73: Just to flag, eventually I suspect this score handling will move somewhere else given we will also want to use it inside Autoemulate.compare(), but no need to decide that here.

radka-j (approved):

autoemulate/experimental/tuner.py
sgreenbury: I think this is for cases when we have any emulators that returns parameters of distribution (e.g. mean and std in [sklearn's GP](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor.predict)). We might consider always returning a `DsitributionLike` from emulators as part of the refactor? (relates to #327)

radka-j: Makes sense, thanks! Since we already have an issue for this let's leave it as is.

PR #388, opened 2025-04-03
Add cross_validate function

### LINKED ISSUE #344: Implement cross-validation approach

radka-j:
Closes #344

This PR:
- adds a `cross_validate` function that is compatible with any of the sklearn cross validation methods (e.g., KFold, LeaveOneOut) and returns an r2 and rmse score per validation fold (the metrics that AutoEmulate implements)
    - also removes `cross_validate` method from the Emulator base class 
- moves some of the validation evaluation code in the Tuner to a new function called `evaluate` as we will want to reuse this in a few places
- updates Tuner docstrings

Questions:
- Currently both of the new functions are in a new file called `model_selection`, does this seem right? 
- Are there args to cross_validate we're missing (e.g., batch_size)?

Somewhere we also need to revisit data handling throughout the codebase (e.g., what should be a Database vs DataLoader and where/how best to handle either) but it might be worth to wait for #371 when we will have an idea all the components and how they talk to each other.


sgreenbury:
I think this looks great! We could merge for the MVP as is I think but adding a few thoughts/code for either here or subsequent work on this:
- I think the API being free functions in the `model_selection.py` seems like a good option to me. Another option might be to include a method on the `Emulator` for evaluation, e.g.:
  ```python
   def evaluate(self, x: InputLike, y: InputLike, metric: Callable) -> float ...
   ```
   However, it's not clear to me if adding this API to the emulator is useful and this might be better to return to following MVP refactor.
- Using sklearn metrics or possibly [torchmetrics](https://lightning.ai/docs/torchmetrics/stable/) (seems to be currently most widely used metrics API for torch directly?) (explored adding this in #392)
- Regarding the batch size: we could implement a loop over validation batches and collect scores per batch? I added an example of trying this in #392 as well

sgreenbury (approved):

sgreenbury:
Merging this now following review with @radka-j @EdwinB12

Issue #394, opened 2025-04-09
Docs are built when any PR is made. It probabaly should only be triggered when merging to main


EdwinB12:

sgreenbury:
Closed by #395

PR #395, opened 2025-04-09
Make docs only run on pushes and PR to main

EdwinB12:


EdwinB12:
resolve #394

.github/workflows/docs.yaml
```diff
@@ -1,5 +1,13 @@
 name: Docs
-on: [push, pull_request, workflow_dispatch]
+#on: [push, pull_request, workflow_dispatch]
```
sgreenbury, line 2:
```suggestion
```

sgreenbury (approved):

PR #384, opened 2025-04-02
Add Gaussian Process for refactor

sgreenbury:
Closes #349.

Remaining tasks:
- [x] Fix warning: `GPInputWarning: The input matches the stored training data. Did you forget to call model.train()?`
- [x] Check logging functionality
- [x] Check values for R2Score
- [x] Add preprocessor
- [x] Add conversion to tensors in init

PR #399, opened 2025-04-14
Add refactored LGBM model to experimental emulators

### LINKED ISSUE #351: Refactor: initial impl Light GBM

edwardchalstrey1:
- Closes #351 
- Adds LightGBM to experimental, set up to have a `get_tune_config` method
- Refactors the original emulator to not use `scipy.stats` and use `np` instead
- Adds tests for simply running the model predict and running the tuner
- Add a `_convert_to_numpy` function to utils, used by the fit method


autoemulate/experimental/data/utils.py
```diff
@@ -93,6 +93,18 @@ def _convert_to_tensors(
...
         )
 
+    def _convert_to_numpy(
+        self,
+        x: InputLike,
+        y: InputLike | None = None,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Convert InputLike x, y to tuple of numpy arrays.
+        """
```
radka-j, line 103: InputLike can be a numpy array so maybe include an if that just returns x,y in that case?

autoemulate/experimental/data/utils.py
```diff
@@ -93,6 +93,18 @@ def _convert_to_tensors(
...
+        x, y = self._convert_to_tensors(x, y)
```
radka-j, line 104: if y is None, `_convert_to_tensors` returns just an x so we need to handle that in the code

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+from torch import Tensor
+
+from autoemulate.experimental.emulators.base import (
+    Emulator,
+    InputTypeMixin,
+)
+from autoemulate.experimental.types import InputLike, OutputLike
+
+
+class LightGBM(Emulator, InputTypeMixin, BaseEstimator, RegressorMixin):
```
radka-j, line 14: Why are we inheriting from the two sklearn classes here? Can we do without it?

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+        self.subsample = subsample
+        self.colsample_bytree = colsample_bytree
+        self.reg_alpha = reg_alpha
+        self.reg_lambda = reg_lambda
+        self.random_state = random_state
+        self.n_jobs = n_jobs
+        self.importance_type = importance_type
+        self.verbose = verbose
+
+    def fit(self, x: InputLike, y: InputLike | None, sample_weight=None, **kwargs):
```
radka-j, line 63: Why do we have kwargs here, it doesn't look like we use them?

What is sample_weight for? If we need it, can we have it as part of the inputs args rather than in the fit method? Just thinking about having consistent API, in the top level AutoEmulate class we never pass anything to the fit method other than x and y.

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+        """Fits the emulator to the data."""
+
+        x, y = self._convert_to_numpy(x, y)
+
+        self.n_features_in_ = x.shape[1]
+
+        x, y = check_X_y(
```
radka-j, line 70: What does chech_X_y do? I feel like any data checks and conversions we want to do should be in the InputTypeMixin (or something similar)?

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1, line 14: Is there a good reason to also inherit from the sklearn `BaseEstimator` and `RegressorMixin` classes here or should we remove these?

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+            "n_estimators": [np.random.randint(50, 1000)],
+            "reg_alpha": [10 ** np.random.uniform(-3, 0)],
+            "reg_lambda": [10 ** np.random.uniform(-3, 0)],
+        }
+
+    @property
+    def model_name(self):
+        return self.__class__.__name__
+
+    def _more_tags(self):
```
radka-j, line 124: Can we remove this method and just not call it above?

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+            verbose=self.verbose,
+        )
+
+        self.model_.fit(x, y, sample_weight=sample_weight)
+        self.is_fitted_ = True
+
+    def predict(self, x: InputLike) -> OutputLike:
+        """Predicts the output of the emulator for a given input."""
+        x = check_array(x)
+        check_is_fitted(self, "is_fitted_")
```
radka-j, line 102: If using this check function is dependent on this object inheriting the the sklearn base objects I'd be in favour of not doing the inheritance and just getting rid of this (and replacing it with our own check if we think that's necessary)

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,125 @@
...
+            importance_type=self.importance_type,
+            verbose=self.verbose,
+        )
+
+        self.model_.fit(x, y, sample_weight=sample_weight)
+        self.is_fitted_ = True
+
+    def predict(self, x: InputLike) -> OutputLike:
+        """Predicts the output of the emulator for a given input."""
+        x = check_array(x)
```
radka-j, line 101: Similar to comment above, I think we should use our own data conversion/check methods, although could be persuaded we should keep it here.

autoemulate/experimental/tuner.py
```diff
@@ -64,7 +64,12 @@ def run(self, model_class: type[Emulator]) -> tuple[list[float], list[ModelConfi
             }
 
             # TODO: consider whether to pass as tensors or dataloader
-            m = model_class(train_x, train_y, **model_config)
+            # TODO: is there a better way to distinguish between models that
+            # require training data for initialisation as well as fitting?
+            if "x" in model_class.__init__.__code__.co_varnames:
```
radka-j, line 69: Is it possible to do a check based on the different emulator types we've defined (e.g., is it a GP model or not)?

autoemulate/experimental/emulators/lightgbm.py
radka-j: Yes we should remove this. It's because in the original AutoEmulate implementation everything was made to look like an sklearn object but that is not necessary anymore.

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: Good points:
1. **kwargs are unused so will remove
2. sample_weight defaults to None in the actual LGBM model fit method and isn't changed anywhere in AutoEmulate so I think also fine to remove

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: This checks X and y have the same length and also that X to be 2D and y 1D, however there are also other checks you can include that in the old Autoemulate have been set up differently depending on the emulator class e.g. for SVM:

```
X, y = check_X_y(
            X,
            y,
            multi_output=self._more_tags()["multioutput"],
            y_numeric=True,
            ensure_min_samples=2,
        )
 ```
 
 and for polynomials:
 
 ```
X, y = check_X_y(X, y, multi_output=True, y_numeric=True, dtype=np.float64)
```

Perhaps for now it's best to leave these checks in the fit methods in, then to refactor into the InputTypeMixin later if several of the models we include in experimental are all doing basically the same check or we disagree with some of the differences and think that a simple dimensionality check is sufficient?

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: Yes, this seems like a weird way to do it - `multi_output` defaults to False so in this case can remove completely

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: I think this might be fine to keep, it just checks that it's a non-empty 2D array containing only finite values. The function is imported and doesn't rely on the sklearn base class i.e. `from sklearn.utils.validation import check_array`

autoemulate/experimental/tuner.py
edwardchalstrey1: Initially thumbed up, but actually I feel like your way would need to be updated when new emulators are added, but my way is more flexible, what do you think?

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -0,0 +1,119 @@
...
+            "max_depth": [np.random.randint(-1, 12)],
+            "learning_rate": [10 ** np.random.uniform(-3, -1)],
+            "n_estimators": [np.random.randint(50, 1000)],
+            "reg_alpha": [10 ** np.random.uniform(-3, 0)],
+            "reg_lambda": [10 ** np.random.uniform(-3, 0)],
+        }
+
+    @property
+    def model_name(self):
+        return self.__class__.__name__
```
sgreenbury, line 119: Currently we've not been using a `model_name` in the refactor though I think for displaying results or for some of the input config interface (e.g. setting which models) it will be useful.

I think if we're including this method here it would be worth adding to the `Emulator` base class as a default implementation. And possibly as a `@classmethod` so it can be called on the class rather than an instance as some of the function interfaces work with classes as opposed to instances.

autoemulate/experimental/tuner.py
sgreenbury: We might prefer to not handle this at the caller level (i.e. instantiating the class) but instead within each class's `__init__`? With the aim that the init interface can be the same for all emulators.

For cases like LightGBM where the training data is not required for the init, `x` and `y` can be unused or e.g. `_, _ = x, y`.

In #384 ([code](https://github.com/alan-turing-institute/autoemulate/blob/5aeee8d7e23457a7b61a4dd0a678dc527808067d/autoemulate/experimental/emulators/base.py#L23-L25)), the base class's `__init__` added x, y for this purpose. This approach might need to change if we update the interfaces as part of #400 (e.g. with dataloaders instead of x, y).

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: @sgreenbury Thanks, this works well - I have moved to the base class and added a test for the LGBM case

autoemulate/experimental/emulators/lightgbm.py
edwardchalstrey1: #419

autoemulate/experimental/data/utils.py
```diff
@@ -93,6 +93,26 @@ def _convert_to_tensors(
...
+        """
+        Convert InputLike x, y to tuple of numpy arrays.
+        """
+        if isinstance(x, np.ndarray) and (y is None or isinstance(y, np.ndarray)):
+            return x, y
+
+        result = self._convert_to_tensors(x, y)
+        if isinstance(result, tuple):
+            x, y = result
+            if y.ndim > 1:  # _convert_to_tensors may return 2D y
```
edwardchalstrey1, line 110: @sgreenbury @radka-j I noticed that the reason we need to adjust the dimensionality here is actually because `_convert_to_numpy` includes a call to `_convert_to_tensors` which we included to avoid duplicating code. Do you think this handles it, or would it be better to refactor `_convert_to_numpy` to not use `_convert_to_tensors`?

edwardchalstrey1: Or alternatively, do we anticipate there being some cases where y *should* be a 2D numpy array?

radka-j (approved):
🚀

Issue #373, opened 2025-03-28
Refactor: create base Simulator class


radka-j:
Let's use this issue to figure out what the base Simulator class should look like.

Once that is agreed, refactor existing simulations in the codebase to be instances of this class to make them compatible with the active learners.

For reference: we have an initial implementation [here](https://github.com/alan-turing-institute/autoemulate/blob/fa9ae9147840be9141955d86222550b0ecdb1386/autoemulate/experimental/learners/base.py#L223). Eventually this should be moved to the `simulations` directory.

marjanfamili:
Do we need to make considerations  for time dependant outputs ?

radka-j:
I think it's OK to start with a base simulator without time as we have now and come back to this once we are also thinking about how best to handle time when training emulators.

cisprague:
Yes, something like that. Also as mentioned in https://github.com/alan-turing-institute/autoemulate/issues/345#issuecomment-2805279740, maybe it would be good to use underscore methods for subclassers to implement and non-underscored for methods that use those implemented methods with checks.

radka-j:
Thanks for sharing that @marjanfamili! I think your `run_simulation` is `sample_forward` and `generate_initial_samples` is `sample_inputs` in the example based on @cisprague base class so it will be easy enough to reconcile the two and include the properties you need as well.  

I think we should put it in:

```
from autoemulate.experimetal.simulations.base import BaseSimulator
```

radka-j:
Closed in #409

PR #409, opened 2025-04-22
abstract base simulator

marjanfamili:
#373 
Abstract base simulator class 

```python 
from autoemulate.simulations.base_simulator import BaseSimulator
``` 
- [x] add test - > mock simulator with the base class and then end to end test 
- [x] fix pre-commit 
- [x] add base simulator


radka-j:
Fo the sake of starting a discussion, here's what I imagined we will do *eventually*:
- [ ] refactor @cisprague active learning code do that the Base class becomes a mixin class
- [ ] after that refactor, move @cisprague base `Simulator` to `autoemulate.experimnetal.simulations.base` (this now uses the mixin class instead of inheriting from `Base`)
- [ ] add the properties/methods that @marjanfamili requires that are currently missing (e.g., `sample_inputs`, and `param_names` and `output_names`)
- [ ] update @marjanfamili work to use these method names (e.g., `sample` instead of `run_simulation`)

I am happy for this base simulator to go into the main codebase and for that refactor to happen at a later stage. But I thought it might be worth flagging here that this might happen after (and of course everyone feel free to tell me otherwise).

sgreenbury:
> Fo the sake of starting a discussion, here's what I imagined we will do _eventually_:
> 
>     * [ ]  refactor @cisprague active learning code do that the Base class becomes a mixin class
> 
>     * [ ]  after that refactor, move @cisprague base `Simulator` to `autoemulate.experimnetal.simulations.base` (this now uses the mixin class instead of inheriting from `Base`)
> 
>     * [ ]  add the properties/methods that @marjanfamili requires that are currently missing (e.g., `sample_inputs`, and `param_names` and `output_names`)
> 
>     * [ ]  update @marjanfamili work to use these method names (e.g., `sample` instead of `run_simulation`)

This order for refactoring sounds good to me! Perhaps we could add wider integration of the base class across all examples as a fifth item given this is mentioned in #373. And potentially a separate issue could be opened for updating the examples to demonstrate using active learning approaches facilitated by the refactor.

> I am happy for this base simulator to go into the main codebase and for that refactor to happen at a later stage. But I thought it might be worth flagging here that this might happen after (and of course everyone feel free to tell me otherwise).

Agreed, this will change further given the points you mentioned above. Also happy for this to go to main - great to have the test with `MockSimulator` and the two `Simulator` abstract base classes on the same branch for the refactor @radka-j mentions in points 2 and 3 above.

For this PR, perhaps we can:
- move the `autoemulate/simulations/base_simulator.py` to`autoemulate/experimental/simulations/base.py` (since the refactor updates are being added there for now)
- rename the base class to `Simulator` since the base emulator is called `Emulator` and this would be analogous

marjanfamili:
I think having the word `Base` in the name is suggestive of the function, so could also change the name of `Emulator` to `BaseEmulator`

sgreenbury:
Following discussion with @marjanfamili, the aim for this PR is to:
- add initial simulator impl (#373) based on example application aiming to support #412 
- the base class is ok to be not in experimental here but will be as part of combining #414

sgreenbury (approved):
Thanks @marjanfamili, looks good to me to merge with further modifications planned for #414

PR #411, opened 2025-04-22
Refactor active learning

### LINKED ISSUE #345: Refactor: Active learning following refactor
This issue is to capture implementing active learning within the codebase following initial refactor. Relates to #301, #316.

sgreenbury:
Closes #345.


autoemulate/experimental/data/validation.py
```diff
@@ -0,0 +1,224 @@
+import torch
+from autoemulate.experimental.types import InputLike, OutputLike, TensorLike
+
+
+# TODO: consider reverting but currrently not all emulators are hashable
+# (e.g. GaussianProcessExact)
+# @dataclass(kw_only=True)
```
sgreenbury, line 7: Remove comment as all staticmethods

docs/tutorials/04_active_learning.ipynb
radka-j, line 1: Code cell 6 still has the old GP implementation, need to remove this.

autoemulate/experimental/learners/base.py
```diff
@@ -486,9 +213,16 @@ def fit(self, *args):
...
-        self.metrics["trace"].append(self.trace(Sigma, self.out_dim).item())
-        self.metrics["logdet"].append(self.logdet(Sigma, self.out_dim).item())
-        self.metrics["max_eigval"].append(self.max_eigval(Sigma).item())
+
+        # If Gaussian output
+        if isinstance(output, MultivariateNormal):
+            assert isinstance(output.variance, TensorLike)
+            assert output.variance.ndim == 2
+            assert output.variance.shape[1] == self.out_dim
+            std = torch.sqrt(output.variance)
```
sgreenbury, line 222: Revised to use get the standard deviation here aiming to be consistent with previous GP impl used here.

sgreenbury:
Following [c908238](https://github.com/alan-turing-institute/autoemulate/pull/411/commits/c908238e390b933ea9f54151bd7016ed893dab46) and a few small changes, the notebook has been re-run now. There remain a couple of issues I have seen so far:
- For sin simulator, the `Adaptive_D_optimal` does not appear to converge with the other results as expected
- For the projectile simulator:
  - logdet does not decrease as expected
  - `Adaptive_D_optimal` does not run successfully

Since the change has been to the GP class, I think the explanation will be differences there. These might include:
- normalizing y (initial test of previous version without normalization might explain some of the difference)
- callbacks (early stopping and LR scheduler are present in non-experimental)

autoemulate/experimental/learners/base.py
sgreenbury: Revised to use `.variance` property of `MultivariateNormal` (which is the expected diagonal of the covariance matrix). Check this as part of #430 and has implications for #434.

sgreenbury:
Adding this comparison we discussed:
normalized_and_unnormalized_output_comparison

The left-hand plot shows the unnormalized case where the increasing log determinant is due to the GP learning the loc/scale of the output while the right-hand plot the scale is loc=0 scale=1 so that the scale of the uncertainty captured by the log det is captured from the start.

sgreenbury:
Thanks for the review and discussion @radka-j and @cisprague - merging this now as discussed.

Issue #348, opened 2025-03-19
Refactor: preprocessing and dimensionality reduction


ContiPaolo:
Dimensionality reduction (for both inputs and outputs) is essential for handling large-scale spatio-temporal data. This requires training and, if using deep learning, hyper-parameter (HP) tuning.

We need to define how to handle this in the refactoring process. There are three possible approaches:

1. **Embedded in the emulator**: Dimensionality reduction and emulator training happen together. More general but computationally expensive.
2. **Separate pre-step:** Train dimensionality reduction first, then train the emulator on reduced data. More efficient but might lead to a suboptimal combination of reduction + emulator.
3. **Hybrid approach (recommended):** First, train dimensionality reduction and compare emulator performance. Then, retrain both together to fine-tune the system just on the best selected model.

In terms of refactoring, we might choose an approach (maybe the hybrid) as default, but I think we should flexibility to switch between methods. This might require making the preprocessing step a trainable component that can be:
- Trained together with the emulator (approach 1),
- Trained separately beforehand (approach 2),
- Retrained after emulator selection (approach 3).

sgreenbury:
Thanks @ContiPaolo for adding this, this is very helpful and I like the approaches outlined! Relates to #323 too - it seems like the options might be also generally applicable to any processing that is trainable (as opposed to some that might be single transformations of the data, e.g. z-score).

For approach 1/3, I'm wondering how we flexibly integrate different dimensionality reduction/trainable processing approaches with emulators. Perhaps creating `nn.Sequential` model with an emulator and processor that both implement the PyTorch backend so that they can be fit with the same API could work. A custom loss might also be required in implementing to also optimise the processing aspect. Being able to compose different processing models and emulators at runtime could be a nice feature too to allow searching over different model compositions.

radka-j:
We agreed that we will handle distribution transforms as a separate issue. There already is one with notes on potential approaches in #376.

sgreenbury:
Thanks @radka-j for connecting to #376, adding some ideas for the specific tasks to cover here for this issue (and we might consider revising some of these further into sub-issues too):
- [ ] Add AutoEmulate `Transform` class
- [ ] Add AutoEmulate `TransformModule` (also an `nn.Module`) (this could be a separate issue)
- [ ] Implement the transform for PCA and VAE in current main
- [ ] Add classes for emulator
  - [ ] With fixed preprocessing and postprocessing
[...]

PR #415, opened 2025-04-22
Refactor sklearn models for experimental

### LINKED ISSUE #405: Refactor: Create SklearnBackend class inheriting from Emulator base class & add first example (svm)
### Summary

- See diagram in #322 for where `SklearnBackend` should sit
- See [this comment](https://github.com/alan-turing-institute/autoemulate/issues/315#issuecomment-2720591494) - idea may be to introduce this functionality but deprecate later

### Basic Example

New SVM class should replicate the existing `SupportVectorMachines` emulator class functionality, but in the new experimental refactor code, inheriting the from a new `SklearnBackend` class which itself inherits from the `Emulator` base class.

### Drawbacks

Retaining sklearn models which might be re-implementable with Pytorch

### Unresolved questions

_No response_

### Implementation PR

_No response_

### Reference Issues

_No response_

edwardchalstrey1:
- Based off #399 - merge this first
- Closes #405 

Questions:
- [x] What are the full set of sklearn models that we want to keep?
    - Just SVM and Random Forest for now 
- [x] Do we actually need the `SklearnBackend`?
- [x] If we do keep `SklearnBackend`, should LightGBM use this?
    - Probably not, if it is sufficiently similar then maybe - have a look at the docs. If we did, should call `SklearnBackend` something different
- [x] Why are the checks in each model fit slightly different and do they need to be? If they can be standardised, we can move them into the fit method of `SklearnBackend` - if not, perhaps keep them in the specific model's fit func
- [x] Add type hints to model args


autoemulate/experimental/emulators/base.py
```diff
@@ -168,3 +168,29 @@ def get_tune_config():
...
+    """
+
+    def __init__(
+        self, x: InputLike | None = None, y: InputLike | None = None, **kwargs
+    ):
+        pass
+
+    def fit(self, x: InputLike, y: InputLike | None):
+        msg = "Subclasses must implement the `fit` method."
+        raise NotImplementedError(msg)
```
sgreenbury, line 188: We might be able to write a generic default implementation for an sklearn model here? Similar to the approach with the [PyTorchBackend](https://github.com/alan-turing-institute/autoemulate/blob/46ca7a6c6b35cdc7cb04e687f89fe8cc921e1f75/autoemulate/experimental/emulators/base.py#L85-L149). For this to work though, I think we'd need to add `.model_` attribute to the backend (or inherit directly from a model class).

autoemulate/experimental/emulators/base.py
```diff
@@ -168,3 +168,29 @@ def get_tune_config():
...
+
+
+class SklearnBackend(Emulator):
+    """
+    SklearnBackend is a sklearn model and implements the base class.
+    This provides default implementations to further subclasses.
+    This means that models can subclass and only need to implement
+    `.fit()` and `.predict()` to have an emulator to be run in `AutoEmulate`
+    """
+
```
sgreenbury, line 180: Could add an attribute here with a type such as: `class SklearnEstimator(BaseEstimator, RegressorMixin): ...`. This could be referenced in default impls of the backend.

```suggestion
    model_: SklearnEstimator
```

autoemulate/experimental/emulators/base.py
edwardchalstrey1: Yes, I think it will make more sense how to do this once I've added a few other models

autoemulate/experimental/emulators/base.py
edwardchalstrey1:
```
class SklearnEstimator(BaseEstimator, RegressorMixin): ...

class SklearnBackend(Emulator):
    """
    SklearnBackend is a sklearn model and implements the base class.
    This provides default implementations to further subclasses.
    This means that models can subclass and only need to implement
    `.fit()` and `.predict()` to have an emulator to be run in `AutoEmulate`
    """

    model: SklearnEstimator

    def __init__(
        self, x: InputLike | None = None, y: InputLike | None = None, **kwargs
    ):
        pass
```

sgreenbury:
Adding the code from discussion with @edwardchalstrey1 and @radka-j on possible revision to the base class (relating to #322):

```python
class Emulator(ABC, InputTypeMixin):
    """
    The interface containing methods on emulators that are
    expected by downstream dependents. This includes:
    - `AutoEmulate`
    """

    @abstractmethod
    def __init__(
        self, x: InputLike | None = None, y: InputLike | None = None, **kwargs
    ): ...

    @classmethod
    def model_name(cls) -> str:
        return cls.__name__

    @abstractmethod
    def _fit(self, x: InputLike, y: InputLike | None): ...


    @abstractmethod
    def check(self, x: InputLike, y: InputLike | None): ...

    @abstractmethod
    def convert(self, x: InputLike, y: InputLike | None) -> tuple[InputLike, InputLike | None]: ...

    def fit(self, x, y):
        """
        Fit the model to the data.

        Parameters
        ----------
            x: InputLike
                Input features as numpy array, PyTorch tensor, or DataLoader.
            y: OutputLike or None
                Target values (not needed if x is a DataLoader).

        Returns
        -------
            None
        """
        x, y = self.convert(x, y)
        self.check(x, y)
        self._fit(x, y)

    @abstractmethod
    def predict(self, x: InputLike) -> OutputLike: ...

    @staticmethod
    @abstractmethod
    def get_tune_config() -> TuneConfig: ...

```

edwardchalstrey1:
Models to inclide:
- [ ] SVM
- [ ] Random Forest

autoemulate/experimental/emulators/randomforest.py
```diff
@@ -0,0 +1,79 @@
...
+            bootstrap=self.bootstrap,
+            random_state=self.random_state,
+        )
+
+    def fit(self, x: TensorLike, y: TensorLike | None):
+        """Fits the emulator to the data."""
+        x, y = self._convert_to_numpy(x, y)
+        self.n_features_in_ = x.shape[1]
+
+        # y = y.ravel()  # Ensure y is 1-dimensional
```
edwardchalstrey1, line 57: Doing this would break the 2d test case, however not doing it results in the following warning even on the 1D test (caused by our dimensionality handling in the `_convert_...` functions:

```
A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples,), for example using ravel()
```

This is probably fine though

edwardchalstrey1:
### TODOs week 6th May

- [ ] Uncomment the `is_multioutput()` func in `base.py` and ensure this doesn't break sklearn models
- [ ] Address pre-commit issues
- [ ] Decide whether to work on #422 here or to merge this PR first

autoemulate/experimental/data/utils.py
```diff
@@ -165,6 +171,18 @@ def _random_split(
...
+    def _normalize(x: TensorLike) -> tuple[TensorLike, TensorLike, TensorLike]:
+        x_mean = x.mean(0, keepdim=True)
+        x_std = x.std(0, keepdim=True)
+        return (x - x_mean) / x_std, x_mean, x_std
+
+    @staticmethod
+    def _denormalize(
+        x: TensorLike, x_mean: TensorLike, x_std: TensorLike
+    ) -> TensorLike:
+        return (x * x_std) + x_mean
```
sgreenbury, line 184: Added this here as initial functionality form preprocessing outputs. This would be good to revisit in #348 where this might be better as distinct functionality that can be combined with the model and/or updated with the API used there.

sgreenbury (approved):
As discussed, this looks good to merge with subsequent changes with checks/validation to follow in #422, thanks @edwardchalstrey1!

autoemulate/experimental/data/utils.py
sgreenbury: Also relates to #437

PR #431, opened 2025-04-28
Fix error raised in sensitivity analysis docs

### LINKED ISSUE #355: Sensitivity analysis docs example gives error
The sensitivity analysis in the docs [here](https://alan-turing-institute.github.io/autoemulate/getting-started/quickstart.html#sensitivity-analysis) gives an error, this should be fixed.

I would also be in favour of moving this from the QuickStart tutorial to the `A deeper dive into AutoEmulate` section. This section ends with "Downstream tasks" examples, which currently only has History matching. I think we should have sensitivity analysis here as well.

sgreenbury:
Closes #355.


radka-j (approved):
Sensitivity analysis in notebooks runs fine for me locally.

PR #446, opened 2025-04-30
Update docs to not execute workflow notebook

radka-j:
One of the cells in the workflow notebook takes too long to run so it gets interrupted during the build, meaning the rest of the notebook doesn't render. I am trying to set up the docs so that the workflow notebook doesn't execute on each build (and instead the pre-rendered content should be displayed). The updates here I based on suggestions from ChatGPT (which has gotten it wrong before). Any suggestions for better ways to go about this welcome!


docs/tutorials/07_AE_workflow.ipynb
```diff
@@ -464,6 +464,9 @@
   }
  ],
  "metadata": {
+  "jupyter": {
+   "execute_notebook": "off"
```
radka-j, line 468:
```suggestion
   "tags": ["skip-execution"]
```

sgreenbury:
Now updated to not execute notebook and include outputs in notebooks + fix the repeated displays in dashboard. Merging now with additional changes #450.

PR #456, opened 2025-05-01
400 revise input types

EdwinB12:
All input types in the emulator methods are changed from InputLike to TensorLike. 

After some discussion with @sgreenbury - We have decided to keep other input types as InputLike for other modules such as model_selection.py and validation.py

### DIFF
--- a/autoemulate/experimental/tuner.py
+++ b/autoemulate/experimental/tuner.py
@@ -1,10 +1,11 @@
 import numpy as np
+from torch.utils.data import DataLoader, Dataset
 from torchmetrics import R2Score
 from tqdm import tqdm
 
 from autoemulate.experimental.emulators.base import Emulator, InputTypeMixin
 from autoemulate.experimental.model_selection import evaluate
-from autoemulate.experimental.types import InputLike, ModelConfig, TuneConfig
+from autoemulate.experimental.types import ModelConfig, TensorLike, TuneConfig
 
 
 class Tuner(InputTypeMixin):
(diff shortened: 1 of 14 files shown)


edwardchalstrey1 (commented):
Not a strongly held opinion, but is there much need for us to have created `TensorLike` as oppose to using `torch.Tensor` directly?

autoemulate/experimental/tuner.py
```diff
@@ -13,19 +14,19 @@ class Tuner(InputTypeMixin):
 
     Parameters
     ----------
-    X: InputLike
-        Input features as numpy array, PyTorch tensor, or Dataset.
+    X: DataLoader | Dataset
+        Input features as DataLoader or Dataset.
     y: OutputLine or None
```
edwardchalstrey1, line 19:
```suggestion
    y: OutputLike or None
```

edwardchalstrey1 (approved):

sgreenbury:
> Not a strongly held opinion, but is there much need for us to have created `TensorLike` as oppose to using `torch.Tensor` directly?

Yes it is just a 1:1 mapping here (unlike `InputLike`) though still useful to have alongside other aliases in `types.py` that provide a package-level definition of our types for use in the APIs/impl. It could also simplify refactoring of future changes to types in the future since the alias could be updated.

sgreenbury:
Thanks @EdwinB12 and @edwardchalstrey1 for the review discussion, adding further comments here and will go ahead and merge:
- Updating the Tuner API to `Dataset | DataLoader` had the downside of making the APIs more complicated across the classes `AutoEmulate`, `Tuner`, `Emulator` - we discussed that for now this would be ok to revert to `InputLike` so that `AutoEmulate` and `Tuner` both have the same input API (reverted in [b3d2aab](https://github.com/alan-turing-institute/autoemulate/pull/456/commits/b3d2aab233489c616e07839b1944a58d77f5c53d))
- An alternative would be to refactor the `Tuner` API to `TensorLike` but this is has less impact than updating the API at the `Emulator` level since we do not expect to subclass the `Tuner`
- Relatedly the idea of adding further methods to the `Emulator` to facilitate tuning was discussed with #457 being opened to consider this further (e.g. would emulator-specific tuner methods be useful?)

Issue #366, opened 2025-03-27
Should we replace Poetry?


radka-j:
What is the best alternative?

sgreenbury:
[uv](https://github.com/astral-sh/uv) might be a good alternative since it is similar but faster. There is also a [tool for migration](https://github.com/mkniewallner/migrate-to-uv).

EdwinB12:
I think we should try and get this into the new release at end of month, agreed?

sgreenbury:
Adding a note that we can update in the CI too with https://docs.astral.sh/uv/guides/integration/github/. There is also currently a specific version [torch installed in the CI](https://github.com/alan-turing-institute/autoemulate/blob/46ca7a6c6b35cdc7cb04e687f89fe8cc921e1f75/.github/workflows/ci.yaml#L51-L55) - there are [docs on this in uv too](https://docs.astral.sh/uv/guides/integration/pytorch/#using-a-pytorch-index).

edwardchalstrey1:
Note: came up in chat today that Poetry might make it hard, but not impossible to use AutoEmulate on Baskerville

PR #458, opened 2025-05-07
Add device handling

### LINKED ISSUE #455: Refactor: Add a way of setting the device in AutoEmulate

sgreenbury:
Closes #455.

This PR includes:
- Functions and mixin class (`TorchDeviceMixin`) for device handling
- API to set device upon emulator construction
- Extending the `Emulator` base to inherit from `TorchDeviceMixin`
  - Enables data passed to `__init__()`, `fit()`, `predict()` methods to be on any device with the data moved to the expected device within their implementations
  - Assumes that data passed to `forward()` will already move on the expected device
- Updates the experimental emulators to allow device to be passed and the `Tuner` and `AutoEmulate` classes.
- Add tests


autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -60,6 +62,7 @@ def __init__(  # noqa: PLR0913 allow too many arguments since all currently requ
         self.n_jobs = n_jobs
         self.importance_type = importance_type
         self.verbose = verbose
+        TorchDeviceMixin.__init__(self, device, cpu_only=True)
```
edwardchalstrey1, line 65: Nice, this seems neat!

edwardchalstrey1:
Not sure if this is helpful but if so I can open an issue, looks like GitHub for Team/Enterprise subscriptions (does Turing have this?) allows [GPU machines for GitHub actions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#larger-runners) so just wandering if there are tests we want to run for GPU models where device is set to Cuda or something like that

sgreenbury:
> Not sure if this is helpful but if so I can open an issue, looks like GitHub for Team/Enterprise subscriptions (does Turing have this?) allows [GPU machines for GitHub actions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#larger-runners) so just wandering if there are tests we want to run for GPU models where device is set to Cuda or something like that

I really like this idea since it would be great to be able to test GPU support in the CI. I think we might need a separate workflow for it since the current one installs a CPU version of torch I think. This is also discussed in #366.

autoemulate/experimental/emulators/lightgbm.py
sgreenbury: Thanks!

edwardchalstrey1 (approved):
As discussed on Friday, this PR is ready to merge once tests pass, which it looks like they do

PR #464, opened 2025-05-13
Refactor/422 - consistent Utils for experimental code

### LINKED ISSUE #422: Refactor: Add validation functionality and class
From discussion with @cisprague, @edwardchalstrey1, @radka-j, @sgreenbury: following #345 and #399 , revisit the `Base` class to consider where different validation functionality should be. This may involve a hierarchy of validation classes to reflect the emulator, simulator, learner functionality.

edwardchalstrey1:
Closes #422 

This PR refactors commonly used validations, checks and processing into a single `utils.py` and adds tests for each of these functions.

### Reviewers

Read some of my comments below


autoemulate/experimental/data/utils.py
```diff
@@ -110,6 +110,9 @@ def _convert_to_tensors(
                 return x.to(dtype)
             msg = "Number of tensors returned must be greater than zero."
             raise ValueError(msg)
+        # TODO: this error message will never be raised because of
+        # of the value error in _convert_to_dataset
+        # If I remove it, Ruff complains that a return statement is missing
         raise ValueError(
```
edwardchalstrey1, line 116: This error message will never be raised because of the value error in `_convert_to_dataset`, however, if I remove it, Ruff complains that a return statement is missing

autoemulate/experimental/emulators/base.py
```diff
@@ -3,28 +3,32 @@
...
 
+    is_fitted_: bool = False
+
     @abstractmethod
     def _fit(self, x: TensorLike, y: TensorLike): ...
 
     def fit(self, x: TensorLike, y: TensorLike):
         self._check(x, y)
         self._fit(x, y)
+        self.is_fitted_ = True
```
edwardchalstrey1, line 31: I have made this common to all emulators

tests/experimental/test_experimental_utils.py
```diff
@@ -0,0 +1,508 @@
...
+        X = torch.tensor([[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]])
+        y = torch.tensor([1.0, 2.0, 3.0])
+        dataset = TensorDataset(X, y)
+        subset = Subset(dataset, [0, 1])
+        x_tensor, y_tensor = self.mixin._convert_to_tensors(subset)
+
+        assert torch.equal(x_tensor, torch.tensor([[1.0, 2.0], [3.0, 4.0]]))
+        assert torch.equal(y_tensor, torch.tensor([[1.0], [2.0]]))
+
+    # def test_convert_to_tensors_invalid(self):
```
edwardchalstrey1, line 83: See [above comment](https://github.com/alan-turing-institute/autoemulate/pull/464#discussion_r2095306906) why this is commented out for now

tests/experimental/test_experimental_utils.py
```diff
@@ -0,0 +1,508 @@
...
+        """
+        Test check_covariance with an invalid covariance matrix.
+        """
+        sigma = torch.tensor([[1.0, 2.0]])
+        with pytest.raises(ValueError, match="Invalid covariance matrix shape"):
+            self.mixin.check_covariance(tensor_2d, sigma)
+
+    def test_trace(self, sigma_full, tensor_2d, tensor_1d):
+        """
+        Test trace computation for covariance matrices.
```
edwardchalstrey1, line 414: @cisprague I've had a go at adding tests for most of the Validation functions, but I wasn't sure on the `trace` and `logdet` functions - it may be worth you having a review of how I've set all of these up

autoemulate/experimental/data/utils.py
```diff
@@ -183,12 +183,333 @@ def _denormalize(
...
+    Mixin class for validation methods.
+    This class provides static methods for checking the types and shapes of
+    input and output data, as well as validating specific tensor shapes.
+    """
+
+    @staticmethod
+    def _check(x: TensorLike, y: TensorLike | None):
+        """
+        Check the types and shape are correct for the input data.
+        Checks are equivalent to sklearn's check_array.
```
edwardchalstrey1, line 198: The below code attempts to do what sklearn was doing with `check_array`, but considering that we have tensors as oppose to numpy arrays - I'm open to people suggesting this isn't needed or should be changed

radka-j: I think this looks great and captures what we expect for x and y.

autoemulate/experimental/data/utils.py
```diff
@@ -183,12 +183,333 @@ def _denormalize(
...
+
+        return x, y
+
+    @staticmethod
+    def _check_output(output: OutputLike):
+        """
+        Check the types and shape are correct
+        for the output data.
+        """
+        # TODO: add any common checks for output data
```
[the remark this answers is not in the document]
radka-j: I guess this ensures that once we are add new emulators they are not returning anything unexpected. That can be useful but I am also happy to skip it.

radka-j (approved):
Looks great!

autoemulate/experimental/data/utils.py
```diff
@@ -1,13 +1,13 @@
...
-import torch.utils
-import torch.utils.data
-from autoemulate.experimental.types import InputLike, TensorLike
+from autoemulate.experimental.types import InputLike, OutputLike, TensorLike
 from sklearn.utils.validation import check_X_y
 from torch.utils.data import DataLoader, Dataset, Subset, TensorDataset, random_split
 
 
-class InputTypeMixin:
+class ConversionMixin:
```
radka-j, line 10: Minor question. This class has `_normalize` and `_denormalize` methods, which are used to standardise the Ys whereas the `Standardizer` is used for the same purpose but on the Xs. I think it makes sense to have both but maybe we can document a bit better why we have both? Should we be doing something else?

edwardchalstrey1: Good spot, I think these should probably be part of the same Mixin class

edwardchalstrey1: Good point, perhaps we don't need both or perhaps they should be part of the same mixin?
For now I have refactored `ConversionMixin._normalize` to use `Standardizer`.

@sgreenbury perhaps we can discuss the logic around this in the Standardizer init:

```
# Set small values in std as 1.0 instead, see:
# https://github.com/scikit-learn/scikit-learn/blob/812ff67e6725a8ca207a37f5ed4bfeafc5d1265d/sklearn/preprocessing/_data.py#L111
std[std < 10 * torch.finfo(std.dtype).eps] = 1.0
```

sgreenbury: > @sgreenbury perhaps we can discuss the logic around this in the Standardizer init:
> 
> ```
> # Set small values in std as 1.0 instead, see:
> # https://github.com/scikit-learn/scikit-learn/blob/812ff67e6725a8ca207a37f5ed4bfeafc5d1265d/sklearn/preprocessing/_data.py#L111
> std[std < 10 * torch.finfo(std.dtype).eps] = 1.0
> ```

Yes, so this is to not scale elements that are very small as would be dividing by 0.

sgreenbury: I think it might be worth keeping the `Preprocessor`ABC and `Standardizer` in a separate module since this relates to different functionality. They might also be removed following #348.

autoemulate/experimental/data/utils.py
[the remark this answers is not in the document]
sgreenbury: Agree it might be worth checking the following:
- `isinstance(output, OutputLike)` and raise error if not?
- If it is `TensorLike` that it is 2D (we will need to revise this as expand to models with different shape, but could be useful for now)
what do you think?

autoemulate/experimental/data/utils.py
```diff
@@ -175,20 +178,340 @@ def _random_split(
...
+            raise ValueError(f"Expected x to be TensorLike, got {type(x)}")
+
+        if y is not None and not isinstance(y, TensorLike):
+            raise ValueError(f"Expected y to be TensorLike, got {type(y)}")
+
+        # Check x
+        if not torch.isfinite(x).all():
+            msg = "Input tensor x contains non-finite values"
+            raise ValueError(msg)
+        if x.dtype not in (torch.float32, torch.float64, torch.int32, torch.int64):
```
sgreenbury, line 215: Perhaps it's worth adding this union of types in the `types.py` e.g. as `TorchScalarLike`?

autoemulate/experimental/emulators/lightgbm.py
```diff
@@ -112,17 +80,26 @@ def _fit(self, x: TensorLike, y: TensorLike):
...
 
     def _predict(self, x: TensorLike) -> OutputLike:
         """Predicts the output of the emulator for a given input."""
-        x = check_array(x)
-        # TODO (#422): move to predict() and consider if required
-        check_is_fitted(self, "is_fitted_")
         y_pred = self.model_.predict(x)
-        # Ensure the output is a 2D tensor array with shape (n_samples, 1)
-        return Tensor(y_pred.reshape(-1, 1))  # type: ignore PGH003
+        _, y = self._convert_to_tensors(x, y_pred)  # type: ignore PGH003
```
sgreenbury, line 101: Looking at this again, it seems that it is `spmatrix` (`from scipy.sparse import spamtrix`) and `list` return possibilities we don't suppport. Perhaps instead of ignore we can add?
```python
assert not isinstance(y_pred, (spmatrix, list))
```

autoemulate/experimental/emulators/neural_processes/conditional_neural_process.py
```diff
@@ -447,8 +440,8 @@ def _predict(self, x: TensorLike) -> DistributionLike:
             raise ValueError(msg)
 
         # Unsqueeze the batch dimension for x_train, y_train and x_target
-        x_train = self.x_train.unsqueeze(0)
-        y_train = self.y_train.unsqueeze(0)
+        x_train = self.x_train.unsqueeze(0)  # type: ignore PGH003
+        y_train = self.y_train.unsqueeze(0)  # type: ignore PGH003
```
sgreenbury, line 444: Perhaps we can do:
```suggestion
        assert isinstance(self.x_train, TensorLike)
        assert isinstance(self.y_train, TensorLike)
        x_train = self.x_train.unsqueeze(0)
        y_train = self.y_train.unsqueeze(0)
```
since we have checked that it is fitted and x_train and y_train are assigned during the fit.

autoemulate/experimental/data/utils.py
edwardchalstrey1: Doesn't look to be possible to alias `torch.dtype`s so maybe I could switch this logic to just checking that the dtype is not bool - see list of [dtypes](https://docs.pytorch.org/docs/stable/tensor_attributes.html)

edwardchalstrey1: Or perhaps we don't need this check at all

sgreenbury:
I think this is looking great, thanks @edwardchalstrey1. I left a few small comments above.

A remaining question I had was the extent we expect the subclassing of emulators to handle conversion / check shapes of tensors? I think we're still doing that at the moment in e.g. GPs and I think it might be good to move this logic out of any `._fit()`/ `._predict()` calls and into the `.fit()` / `.predict()` to simplify the subclasses. The logic around this might need further refinement as we extend to support data with additional dimensions but might be worthwhile introducing the handling for this at this point. This could also be opened as a separate issue / PR.

radka-j:
@sgreenbury I like the suggestion for `fit` and `predict` to handle checks and conversions. Agree that this could be a separate PR.

tests/experimental/test_experimental_utils.py
edwardchalstrey1: Leave for future PR

autoemulate/experimental/data/utils.py
edwardchalstrey1: Move this back

autoemulate/experimental/data/utils.py
edwardchalstrey1: # dtype's
TorchScalarDtype = (torch.float32, torch.float64, torch.int32, torch.int64)

edwardchalstrey1:
> I think this is looking great, thanks @edwardchalstrey1. I left a few small comments above.
> 
> A remaining question I had was the extent we expect the subclassing of emulators to handle conversion / check shapes of tensors? I think we're still doing that at the moment in e.g. GPs and I think it might be good to move this logic out of any `._fit()`/ `._predict()` calls and into the `.fit()` / `.predict()` to simplify the subclasses. The logic around this might need further refinement as we extend to support data with additional dimensions but might be worthwhile introducing the handling for this at this point. This could also be opened as a separate issue / PR.

Opened #486

edwardchalstrey1:
Final TODOs:
- [ ] update `_check_output`
- [ ] Move Preprocessor and Standardizer back to a separate module

autoemulate/experimental/data/utils.py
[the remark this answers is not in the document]
edwardchalstrey1: Added issue #487

PR #471, opened 2025-05-15
Improve handling of Gaussians distributions

cisprague:
This pull request is intended to improve the handling of Gaussians in AutoEmulate and standardize the outputs active learners can expects from emulators.

Implemented covariance structures:
- [x] Full covariance
- [x] Block-diagonal
- [x] Diagonal
- [x] Separable 
- [ ] Dirac
- [ ] Empirical
- [ ] Ensemble

### DIFF
--- a/autoemulate/experimental/data/gaussian.py
+++ b/autoemulate/experimental/data/gaussian.py
@@ -1,0 +1,113 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+import torch
+from torch import Tensor
+
+@dataclass
+class Gaussian(ABC):
+    """Abstract base for a Gaussian emulator over n inputs and d outputs."""
+    mean: Tensor  # shape [n, d]
+
+    def __post_init__(self):
+        if not isinstance(self.mean, Tensor) or self.mean.ndim != 2:
+            raise ValueError("`mean` must be a 2D torch.Tensor of shape [n, d]")
+
+    @property
+    def n(self) -> int:
+        return self.mean.shape[0]
+
+    @property
+    def d(self) -> int:
+        return self.mean.shape[1]
+
+    @abstractmethod
+    def logdet(self) -> Tensor:
+        """Log‐determinant of the full covariance."""
+        ...
+
+    @abstractmethod
+    def trace(self) -> Tensor:
+        """Trace of the full covariance."""
+        ...
+
+    @abstractmethod
+    def max_eig(self) -> Tensor:
+        """Largest eigenvalue (spectral norm) of the full covariance."""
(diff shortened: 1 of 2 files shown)


cisprague:
Need to discuss how to handle empirically constructed Gaussians. Currently, we have explicit specialised classes to handle them, e.g. `Empirical_Block_Diagonal`, but we could also append class methods `from_samples` to each class, e.g. `Block_Diagonal.from_samples`. Either way, there needs to be a way to convert the classes to `Dense` to construct ensembles with `Ensemble` --- the method `to_dense` accomplishes this.

Lastly, regarding compatibility with `GPyTorch`: they internally use [LinearOperator](https://github.com/cornellius-gp/linear_operator) to handle different kernel specialisations, which looks good but might be a bit more than we need.

cisprague:
> Need to discuss how to handle empirically constructed Gaussians. Currently, we have explicit specialised classes to handle them, e.g. `Empirical_Block_Diagonal`, but we could also append class methods `from_samples` to each class, e.g. `Block_Diagonal.from_samples`. Either way, there needs to be a way to convert the classes to `Dense` to construct ensembles with `Ensemble` --- the method `to_dense` accomplishes this.
> 
> Lastly, regarding compatibility with `GPyTorch`: they internally use [LinearOperator](https://github.com/cornellius-gp/linear_operator) to handle different kernel specialisations, which looks good but might be a bit more than we need.

We now have `.to_dense()` and `.from_dense()` for most of the structured classes. These methods allow for the aggregation of structured distributions into an `Ensemble` class.

```python
# Anisotropic empirical distribution from k samples at n sampling locations, each with d dimensions
k, n, d = 1000, 50, 3
samples = torch.rand(k, n, d)
dist0 = Empirical(samples)

# Isotropic empirical distribution (set off-diagonal elements to zero)
samples = torch.rand(k, n, d)
dist1 = Diagonal.from_dense(Empirical(samples))

# Just to demonstrate to the .to_dense() method
dist1 = Diagonal.from_dense(dist1.to_dense())

# We can combine them into an ensemble
dist2 = Ensemble([dist0, dist1])
print(dist2.logdet(), dist2.trace(), dist2.max_eig())
```

radka-j:
This is looking great! I just want to confirm we all agree with where this code is meant to sit and how it is meant to be used. My understanding is that it will be within the active learning module and these classes are used to reshape the covariance matrix of the torch distribution returned by AutoEmulate (`GaussianLike`) to enable efficient metrics computation. Is this correct?

radka-j:
It would also help me if we had some examples to illustrate when/how the different covariance structures emerge.

Issue #376, opened 2025-03-28
Transformation of uncertainty with dimensionality reduction


marjanfamili:
Transform uncertainty form the latent to the original space for linear dimensionality reduction: explicit transformation

ContiPaolo:
@sgreenbury Some details about the delta method for dimensionality reduction.

- Assuming the latent variables $Z$ are Gaussian distributed: $Z \sim \mathcal{N}(\mu_Z,\Sigma_Z)$.
- Let $g$ be the decoder function that maps these latent variables to the reconstructed variables $X$.
- The delta method approximates the distribution of $X=g(Z)$ using a first-order Taylor expansion around the mean $\mu_Z$: $X\approx g(\mu_Z)+J_g(Z)(Z-\mu_Z)$, where $J_g(Z)$ is the Jacobian of $g$ with respect to $Z$ evaluated at $\mu_Z$ (this should be straightforward to obtain with automatic differentiation).

The distribution of the reconstructed variables can be approximated as: $X\sim\mathcal{N}\left(g(\mu_Z), J_g(\mu_Z)\Sigma_Z J_g(\mu_Z)\right)^\top$, assuming the non-linearity of the decoder $g$ is not extreme.

Therefore the current method `inverse_transform_std` basically needs to implement $J_g(\mu_Z)\Sigma_Z J_g(\mu_Z)$ where the Jacobian $J_g(\mu_Z)$ is given by automatic differentiation while $g(\mu_Z)$ is the reconstruction of the mean (i.e., 'inverse_transform' of the latent mean).

sgreenbury:
Thanks very much for the details on this @ContiPaolo - that's super helpful and sounds great!

Currently the two methods relating to this transform are:
```python
def _inverse_sample(self, x: GaussianLike, n_samples: int = 100) -> GaussianLike:
def _inverse_gaussian(self, x: GaussianLike) -> GaussianLike:
```
For PCA `_inverse_gaussian()` is [wip impl](https://github.com/alan-turing-institute/autoemulate/blob/a6d2de2797d9872eb9d7aa2896e0a44ecddfe314/autoemulate/experimental/transforms/pca.py#L41-L57) with the linear version of this approach - so I think it would make sense to impl this method for non-linear transforms using the approach you describe.

I'll take a look at whether we can have a default impl in the base for this and the sampling approach.

[...]

PR #474, opened 2025-05-16
Refactor: add transforms and transformed emulator

### LINKED ISSUE #348: Refactor: preprocessing and dimensionality reduction

sgreenbury:
Closes #348 and initial impl of #376.

#### Summary
This PR includes:
- An `AutoEmulateTransform` class subclassing `torch.distributions.Transform` that:
  - Enables sequentially composable transforms for inputs and outputs that can flexibly handle both `TensorLike`, `DistributionLike` and with custom functionality for `GaussianLike` outputs.
  - It is subclassed to enable the inclusion of methods for:
    - Fitting after initialization (and checking if fitted)
    - Including approaches to transforming multivariate normals from a basis matrix that can be specified for the transform. This also relates to the #376 where the delta method is included (the Jacobian is used to provide an approximate linear estimate around the mean when inverting). This manifests in the `expanded_basis_matrix` override in application to the VAE.
    - An inverse sample method to enable the empirical construction of the inverted mean and covariance matrix for a `GaussianLike`
- A utility function for matrices to `make_positive_definite`: this function  adds jitter (up to retries) on the diagonal and when this fails clamps the eigenvalues to be positive (up to retries)
- `AutoEmulateTransforms` implemented for:
	- `PCATransform`, `StandardizeTransform` (around mean and std_dev), `VAETransform` (this copies the implementation from the current VAE used as a transform in the non-experimental module)
- `TransformedEmulator` emulator class (the refactor version of the `AutoEmulatePipeline`) with:
	- An emulator with specification of lists of `transforms` for inputs and `target_transforms` for outputs.
	- Has API to enable choice of inverting a wrapped emulator model that returns a `GaussianLike` analytically / with delta method or through a sampling approach
	- Reverts to using diagonal covariance matrix beyond `max_targets` threshold to ensure computationally tractable
- Integration with compare loop is for #531 - a `.tune()` method may need to be directly added to the `TransformedEmulator` so it has a slightly different API and perhaps shouldn't directly subclass `Emulator`

#### Questions
- The separate`.fit()` method could be removed in favour of fitting upon initialization. We could pass `functools.partial` classes to the `TransformedEmulator` in this case instead of fully initialized ones so that the init method can be performed within the `TransformedEmulator`'s init 
- Should the amount of relaxation for `make_positive_definite` be configurable in the emulator or `TransformedEmulator` API
- Should standardization as a transform be included only for `TransformedEmulators` or should it be also included directly within emulators and replace the preprocessor API
- Should we also subclass `ComposeTransform` for `AutoEmulateComposeTransform` to simplify composing `AutoEmulateTransforms`
- Are there additional ways we could be testing the numerical accuracy of the approaches

---

Remaining tasks:
- [x] Open new issues that are beyond scope of this PR for:
  - [x] Learnable transforms (opened #536)
  - [x] Adding transforms/target transforms to the the compare loop / outputs (update: opened #531)
- [x] Confirm value for tolerance for allclose with delta method compared to sampling method for VAE / identify alternative test


autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -170,7 +170,8 @@ def _fit(self, x: TensorLike, y: TensorLike):
             loss = mll(output, y)
             assert isinstance(loss, torch.Tensor)
             loss = -loss
-            loss.backward()
+            # TODO: consider removing retain_graph=True, added for trasnformed emulator
```
sgreenbury, line 173: It would be good to confirm whether this is still required

autoemulate/experimental/emulators/__init__.py
```diff
@@ -1,13 +1,14 @@
 from .gaussian_process.exact import GaussianProcessExact
 from .lightgbm import LightGBM
-from .neural_processes.conditional_neural_process import CNPModule
+
+# from .neural_processes.conditional_neural_process import CNPModule
```
sgreenbury, line 4: See #501

autoemulate/experimental/transforms/utils.py
```diff
@@ -0,0 +1,36 @@
+import torch
+
+
+# TODO: consider if this function is sufficiently robust for all cases.
+# TODO: compare with linear_operator's psd_safe_cholesky:
+# `from linear_operator.utils.cholesky import psd_safe_cholesky``
+def make_positive_definite(
+    cov, epsilon=1e-6, min_eigval=1e-6, max_tries_epsilon=3, max_tries_min_eigval=5
+):
```
sgreenbury, line 9: Further testing of cases where stricter defaults fail could be worth consideration.

autoemulate/experimental/emulators/gaussian_process/exact.py
sgreenbury: Resolved in c9ba448

autoemulate/experimental/emulators/transformed/base.py
```diff
@@ -0,0 +1,230 @@
...
+                If False, use the diagonal covariance matrix. Default is False.
+            max_targets (int): Maximum number of targets to before switching to
+                sampled predictive distribution and diagonal covariance. Default is 200.
+            **kwargs: Additional keyword arguments to pass to the model constructor.
+
+        """
+        self.transforms = transforms if transforms is not None else []
+        self.target_transforms = (
+            target_transforms if target_transforms is not None else []
+        )
```
sgreenbury, line 71: Equivalent to:
```suggestion
        self.transforms = transforms or []
        self.target_transforms = target_transforms or []
```

sgreenbury:
Following discussion:
- It would be good to double check if implications of using torch `Transform` for a `VAE` since the call and inv methods do not guarantee $f^{-1}(f(x)) = x$
- `make_positive_definite` might be useful to pass params too but we can revisit if there are issues that arise in using
- Subclassing `ComposeTransform` to `AutoEmulateComposeTransform` might benefit the API but this could potentially be considered as part of #531 if it seems useful in the top-level API?

autoemulate/experimental/emulators/transformed/base.py
```diff
@@ -0,0 +1,230 @@
...
+        full_covariance: bool = False,
+        max_targets: int = 200,
+        device: DeviceLike = "cpu",
+        **kwargs,
+    ):
+        """Initialize a transformed emulator.
+
+        Parameters
+        ----------
+            x (TensorLike): Input data tensor.
```
radka-j, line 48: This is super minor and I feel bad about even mentioning but related to #543, we should agree exactly on how to write docstrings and stick to it throughout. 

For example I do:
```
x: TensorLike
    Input data tensor.
```

I don't have a strong preference, so lets pick one and maybe open an issue to make this consistent across the code base (somewhere down the line when someone has some spare time :sweat_smile:).

radka-j (commented):
This is looking really great! 🚀 I only have a few small comments, mostly about adding docstrings :)

radka-j:
One last minor thought I had was about whether we want the input args to the inverse transform methods to be `z` rather than `x` or `y`? It might just be easier to track what's what but I am also more than happy to be told that's wrong.

autoemulate/experimental/exploratory/reaction_diffusion/05_reaction_diffusion_experimental.ipynb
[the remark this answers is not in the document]
sgreenbury: I like the idea of moving the tutorial ones to an experimental location in the docs as these are distinct from the other exploratory notebooks that may not be directly in the docs - I'll move this one to `docs/experimental/`

autoemulate/experimental/emulators/transformed/base.py
sgreenbury: Thanks for mentioning this - I completely agree it would be great to be consistent and also include in the style guide! It looks like the current [CONTRIBUTING.md](https://github.com/alan-turing-institute/autoemulate/blob/b4faf29cf5671559163a6e4aa4d45295f7bdd314/CONTRIBUTING.md) indicates [numpy-style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html) so I could update to these throughout this PR and I can open an issue to specifically revisit this?

autoemulate/experimental/transforms/standardize.py
```diff
@@ -0,0 +1,46 @@
...
+
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+
+    def __init__(self, cache_size=0):
+        Transform.__init__(self, cache_size=cache_size)
+
+    def fit(self, x: TensorLike):
+        # TODO: add checks or shape of mean and std
```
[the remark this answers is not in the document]
sgreenbury: Yes this sounds good - I think the `AutoEmulateTransform`. I'll add this.

autoemulate/experimental/emulators/transformed/base.py
radka-j: Sounds great!

sgreenbury:
> One last minor thought I had was about whether we want the input args to the inverse transform methods to be `z` rather than `x` or `y`? It might just be easier to track what's what but I am also more than happy to be told that's wrong.

Yeah I think this could definitely be clearer, perhaps the following refactor would be useful:
- Since torch's `Transform` refers to `x` in the domain and `y` in the codomain, I think the `AutoEmulateTransform` might be best if consistent with this and update the input API to `x` for call and `y` for inverses
- For the `TransformedEmulator`, since this works with `x` for emulator inputs and `y` for targets, using `z` to refer to inputs in the codomain of the transforms for `x` seems like a good option.
- Some of the method names in `TransformedEmulator` might be worth revisiting too (e.g. some refer to `y` like `_inv_transform_y_mvn` while the y transforms in the attribute are called `target_transforms`) - perhaps we could rename the `y` in method names to `target` for consistency. Or alternativelu `target_transforms` to `y_transforms`.

autoemulate/experimental/transforms/vae.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    VAE transform for dimensionality reduction.
+    """
+
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = False
+    vae: VAE | None = None
+
+    def __init__(  # noqa: PLR0913
```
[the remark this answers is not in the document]
sgreenbury: f9bfc3a1bba4371153818316df08416e25a37058

autoemulate/experimental/transforms/standardize.py
```diff
@@ -0,0 +1,46 @@
...
+
+    This transform is effectively a composition of two AffineTransforms with a
+    translation by the mean and a scaling by the inverse of the standard deviation.
+    """
+
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+
+    def __init__(self, cache_size=0):
```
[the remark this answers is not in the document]
sgreenbury: Removed for VAETransform in f9bfc3a1bba4371153818316df08416e25a37058

autoemulate/experimental/exploratory/reaction_diffusion/05_reaction_diffusion_experimental.ipynb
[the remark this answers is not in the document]
sgreenbury: Updated in 499fb4c64ad924eefb663a0da873835c7484fabf

tests/experimental/test_experimental_transformed.py
```diff
@@ -0,0 +1,229 @@
...
+                PCATransform(n_components=3),
+                VAETransform(latent_dim=2),
+            ],
+        ],
+        [
+            # TODO: PCA/VAE both require StandardizeTransform for numerical stability
+            # e.g. "ValueError: Input tensor y contains non-finite values"
+            # TODO: check error when no target transforms are provided
+            # None,
+            # [StandardizeTransform()],
```
sgreenbury, line 126: As above: https://github.com/alan-turing-institute/autoemulate/pull/474/files#r2164309387

radka-j (approved):
This is beautiful!! The code, variable names and documentation are all very clear. I really like the diagrams in the TransformedEmulator docstring :)

PR #479, opened 2025-05-21
Set random seeds for Emulator models

### LINKED ISSUE #465: Refactor: set random seeds
There already is a function in utils for this [here](https://github.com/alan-turing-institute/autoemulate/blob/8b50621e3018b8d83f2b2bfa7fabbad6c6c14359/autoemulate/utils.py#L360). Something like this should be used in the experimental (probably in the AutoEmulate object).

edwardchalstrey1:
Works on #465 - not finished

# TODO

- [x] Branches from #464 so merge that first
- ~~Ensure all emulators using the `set_random_seed` func~~ some models (LGBM and RandomForest) pass the `random_seed` int into the model params, so presumably you don't need to do both
- [x] Ensure all emulators have tests for checking the predictions are consistent when the `random_seed` is set
- [x] Remove the seed being set separately in `CNPDataset`
    - Still doesn't result in `test_fit_predict_deterministic_with_seed` passing
- [x] May want to add a new issue to tackle later to ensure that tests pass for determinism on GPUs
    - Added to #491  

# Reviewers

<details>
<summary>
Completed
</summary>

@EdwinB12 if you have a better sense than me of how CNP/Gaussian processes work and how random seeds should affect the result would be great to get your review here. See my comments below, but my key questions are:
- ~~Why does `test_fit_predict_deterministic_with_seed` fail for CNP?~~
    - There must be some randomness being introduced here still, but I can't see where
    - **Update:** removing CNP #501 so have commented this out for now
- Why does `test_fit_predict_deterministic_with_seed` always pass regardless of `random_state` for GP?
    - **Note:** I guess this doesn't actually matter

</details>

Should we consider refactoring so the seed is set at an even higher level (see Edwin's [comment below](https://github.com/alan-turing-institute/autoemulate/pull/479#pullrequestreview-2878441998)):
 ```
import autoemulate as ae 
ae.set_random_seed(42)
```

### CODE REVIEW

File: autoemulate/experimental/model_selection.py [hunk 1]
```diff
@@ -97,6 +97,8 @@ def cross_validate(
 
         # fit model
         x, y = next(iter(train_loader))
+        # TODO: should we set random_state in the models?
+        # m = model(x, y, random_state=123, **best_model_config)
```
edwardchalstrey1, line 101: 1. Should we set `random_state` for all the models?
2. If so, may make sense to set a default value in `Emulator` base class rather than here

edwardchalstrey1: The deeper question is, do we enforce determinism for AutoEmulate models, if not, should we? Looks like AutoEmulate v0 does set up cross validation with a random_state, but we currently don't do this in experimental/v1

EdwinB12 (dismissed):
Nice work. I've set some comments! 

Mainly two points: 

- I think changing `random_state` to `random_seed` everywhere would be good as it follows our Pytorchy vibe. 
- CNP has randomness in the CNPDataset class which is why those tests are failing I believe. 

I also wonder whether this functionality should actually be at the top level of the whole package. I.E. 

```
import autoemulate as ae 
ae.set_random_seed(42)
```
Wouldn't this then set random seeds for all operations in the codebase?

radka-j (commented):
I agree with Edwin's suggestion that we should be setting random seeds higher up as well. There are also places where we split the data that need to be handled (in the conversion mixin I think and when using the different CV classes from sklearn in model selection). So we can merge this without closing the issue and add the remaining TODOs there.

radka-j (approved):

Issue #438, opened 2025-04-29
Update how handle failed simulations


marjanfamili:
The simulator is given samples and is giving out outputs, if for a sample , the result of the simulation is not valid, it should remove that sample from sample data frame and notify and return samples and their corresponding results which are valid, otherwise ion our workflow, later when autoemulate is called , it will through an error that the length of samples is not the same as the outputs

radka-j:
One approach is for the simulator to internally track parameters which were successfully simulated. It then returns the successful parameter runs as well as results for each. This way it never returns None.

sgreenbury:
From discussion with @radka-j, noting that this currently only affects the Naghavi simulator does not need prioritisation for 1.0.

PR #485, opened 2025-05-23
414 combine simulator implementations

EdwinB12:
Combining our two seperate Simulation Base Classes


autoemulate/experimental/simulations/base.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+
+    def forward(self, x: TensorLike) -> TensorLike:
+        """
+        Generate samples from input data using the simulator. Combines the
+        abstract method `_forward` with some validation checks.
+
+        Parameters
+        ----------
+        x : TensorLike | dict
```
sgreenbury, line 87:
```suggestion
        x : TensorLike
```

autoemulate/experimental/simulations/base.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Process each sample with progress tracking
+        for i in tqdm(range(len(samples)), desc="Running simulations"):
+            result = self.forward(samples[i : i + 1])
+            if result is not None:
+                results.append(result)
+                successful += 1
+
+        # Report results
+        print(
```
sgreenbury, line 120: Perhaps we could add comments for print statements referring to the logging issue?

sgreenbury:
As discussed, as part of https://github.com/alan-turing-institute/autoemulate/issues/438 we can add a flag that says whether to retry failed simulations to make up the requested N.

EdwinB12:
Giving this some more thoughts and some discussion in #492 - We don't really want the Naghavi simulator to be a dependency and therefore its implementation should be in a notebook rather than the codebase. 

We also agree that a seperate notebook demonstrating how to implement your own simulator would be useful.

EdwinB12:
This PR also needs to address the existing simulators in the code base like the projectile

radka-j:
I think we should revisit the idea of having a `SimulatorMetadata` class. There are instances (e.g., `HistoryMatching`) where we might find it useful to have information about the simulation and this should not require the user having to implement a `Simulator._forward()` method.

autoemulate/experimental/simulations/base.py
```diff
@@ -0,0 +1,164 @@
...
+        # define param names and values
+        self._parameters_range = parameters_range
+        self._param_names = list(self._parameters_range.keys())
+        self._param_bounds = list(self._parameters_range.values())
+
+        # Output configuration
+        self._output_variables = (
+            output_variables if output_variables is not None else []
+        )
+        self._output_names: list[str] = []  # Will be populated after first simulation
```
radka-j, line 43: We should have an option for `_output_names` to be populated without having to run the simulation (I think this is quite specific to how the Naghavi simulator is implemented). This is the kind of information that could go in a `SimulatorMetadata` class if we have one.

radka-j:
Closing as this PR is replaced by #520 and the remaining TODOs are captured in #492

PR #496, opened 2025-05-30
Ensure using get_model with a named model works as expected

### LINKED ISSUE #495: Bug: get_model doesn't work correctly when passing in a name
### Description

Running `AutoEmulate.get_model('<name'>)` does not pick the best model with that model name, as the user might expect.

This leads to some confusion, in the notebook in the screenshot below (07), where we pick a best model (`em.compare()`) but also try to pick a model using `em.get_model('GaussianProcess')` which doesn't seem to pick the model 

### Reproduction steps

```bash
    preprocessing_methods = [{"name" : "PCA", "params" : {"reduced_dim": 2}}]
    em = AutoEmulate()
    em.setup(sample_df, results, models=["gp"], scale_output = True, reduce_dim_output=True, preprocessing_methods=preprocessing_methods)
    em.summarise_cv()
```

You would then expect in this case the model from `em.compare()` and `em.get_model('GaussianProcess')` to be the same, but they aren't.

### Version

0.3

### Screenshots

Image

### OS

_No response_

edwardchalstrey1:
Closes #495

### DIFF
--- a/autoemulate/compare.py
+++ b/autoemulate/compare.py
@@ -473,15 +473,15 @@ class AutoEmulate:
                     means = _sum_cvs({full_name: cv_results[full_name]}, metric)
                     current_score = means.iloc[0][f"{metric}"]
 
-                    # Compare with current best
-                    if (metric == "r2" and current_score > best_score) or (
-                        metric != "r2" and current_score < best_score
-                    ):
-                        best_score = current_score
-                        for model in prep_data["models"]:
-                            if get_model_name(model) == full_name:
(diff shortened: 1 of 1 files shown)


edwardchalstrey1:
Hi @marjanfamili I just made a small update to the logic for when you choose a model based on the name e.g. like: `em.get_model('GaussianProcess')` - please could you review?
See #495 - this was the bug I noticed when running notebook 7

marjanfamili (approved):
This PR ensures that when we try  get_model to get a specific model, the correct model is selected

PR #505, opened 2025-06-04
Mcmc hmc

marjanfamili:
#461 

This PR implements MCMC using pyro and torch and a visualisation dashboard for MCMC 

note that I have duplicated the dashboard for history matching and changed the plots and just updated the controls please let me know if you think this should be changed to a dashboard base and then updated for each section 

- [x] The observation values are not reasonable so the NUTS warmup is getting stuck forever trying to determine a mass matrix and step size, this needs to get addressed by running a single simulation and using the outcome of that as observation


radka-j:
I added an issue for possibly refactoring the two dashboards in a separate PR (#509) but think it's totally fine to leave the duplicated version from history matching here as is for now.

autoemulate/mcmc.py
```diff
@@ -0,0 +1,181 @@
...
+from pyro.infer import MCMC
+from pyro.infer import NUTS
+
+
+class MCMCCalibrator:
+    def __init__(
+        self,
+        emulator,
+        sensitivity_results: pd.DataFrame,
+        observations: Dict[str, Tuple[float, float]],
```
radka-j, line 19: It's still a bit unclear to me where the observation variance comes from. 

Option 1: the user has a single observation and the variance here is a hyperparameter. In that case I think we should make that very explicit in the naming or at least the docs.

Option 2: the user has multiple observations and they are passing summary statistics here. In that case, they should just give us all of the observations to use as that will results in more informative posteriors. In that case they do also need to specify an observation noise hyperparameter. Alternatively, we could give them an option to put a prior over it and infer it alongside the other parameters (e.g., that's what they do [here](https://turinglang.org/docs/tutorials/bayesian-differential-equations/#direct-handling-of-bayesian-estimation-with-turing)).

autoemulate/mcmc.py
```diff
@@ -0,0 +1,181 @@
+from typing import Dict
+from typing import List
+from typing import Tuple
+
+import numpy as np
+import pandas as pd
+import pyro
+import pyro.distributions as dist
+import torch
+from pyro.infer import MCMC
```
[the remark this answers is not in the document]
marjanfamili: Thanks Sam, I will

pyproject.toml
```diff
@@ -28,7 +28,7 @@ salib = "^1.5.1"
 torcheval = "^0.0.7"
 anytree = "^2.12.1"
 torchmetrics = "^1.7.1"
-
+pyro-ppl = "^1.9.1"
```
radka-j, line 31: I tried running the notebook and run into issues because `corner` is not installed.

radka-j:
I run the workflow notebook and it works as expected except for the missing corner import as mentioned in my previous comment.

I have some thoughts on how to refactor the `MCMCCalibrator/`calibrator but we can do that after the workshop as part of the PyTorch refactor where we integrate this with the rest of the `experimental` code base (#521). Mainly:
- allow the user to pass multiple observations and handle correctly
- allow the user to run without passing sensitivity results
- possibly modularise by moving all the functionality that refines parameter bounds given NROY samples into a different class (using history matching for this is a very specific workflow that not everyone uses)

radka-j (approved):

pyproject.toml
sgreenbury: Added in [4c87eda](https://github.com/alan-turing-institute/autoemulate/pull/505/commits/4c87eda789bb92df200cbf2ead7e2950b8297cd2)

PR #507, opened 2025-06-05
small bug in HM

marjanfamili:
- Just fixes a small bug, we pass pred, and var to implausibility calculation 

- and moved the check to see if the new samples are plausible or not to the main run 

- and added a few if condition to deal with when no nroy samples are found. 

- also removed instruction from dashboard as it was taking up space and eventually it is going to get merged with MCMC dashboard and only have one calibration dashboard anyway

### DIFF
--- a/autoemulate/history_matching.py
+++ b/autoemulate/history_matching.py
@@ -169,23 +169,13 @@ class HistoryMatching:
 
         # Need to handle discontinuous NROY spaces
         # i.e., region within min/max bounds is RO
-        valid_samples = np.empty((0, nroy_samples.shape[1]))
-        while len(valid_samples) < n_samples:
-            # Generate candidates
-            candidate_samples = np.random.uniform(
-                min_bounds, max_bounds, size=(n_samples, nroy_samples.shape[1])
-            )
 
-            # Filter valid samples based on implausibility and concatenate
-            implausibility = self.calculate_implausibility(candidate_samples)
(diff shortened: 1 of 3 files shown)


radka-j (commented):
Thanks for finding the bug and fixing it!!  🚀

I really like the idea that we only check implausability on the NROY samples if we have an emulator as otherwise it's too expensive. 

I have one question. In the case that we filter the `candidate_samples` for implausability, we can end up with `candidate_samples.shape[0] <= n_samples`. Should we ensure that `candidate_samples` always has `n_samples` rows as in the original implementation or do we think this is OK?

marjanfamili:
> Thanks for finding the bug and fixing it!!  🚀
> 
> I really like the idea that we only check implausability on the NROY samples if we have an emulator as otherwise it's too expensive.
> 
> I have one question. In the case that we filter the `candidate_samples` for implausability, we can end up with `candidate_samples.shape[0] <= n_samples`. Should we ensure that `candidate_samples` always has `n_samples` rows as in the original implementation or do we think this is OK?

I thinks safe to leave this for now, if we faced any issues we can add this in. I had preservations when adding this, I dont know if it is essential that we force the same number of samples in every wave , or let it naturally sample as many, potentially less in a narrower domain.

radka-j (approved):
Good point about allowing fewer samples if the domain narrows rather than forcing many expensive simulations in that narrow range just because.

PR #511, opened 2025-06-06
adding a forward pass of the simulator for parameters at mid range ob…

marjanfamili:
added a forward pass of the simulator for the parameter mid range to set the observations for further calibration


radka-j:
My only comment is that it would be good to make it clear in the notebook that we are calibrating against simulated "observations" just as a proof of concept/for demo purposes rather than as something to do in practice.

PR #512, opened 2025-06-06
Handle random_seed setting at higher level in package

### LINKED ISSUE #465: Refactor: set random seeds

edwardchalstrey1:
Should close #465 

The idea with these changes is that we should have a robust way to enforce determinism in AutoEmulate in both:
- Models we implement
- Models that are just wrappers

### Changes

- The `set_random_seed` function is part of it's own class, which is currently called `RandomMixin` but we could refactor this to just be a utils function that isn't part of a class
- Made `random_seed` a parameter of the `AutoEmulate` class. By default, this is not set, but it can be set to ensure determinism in non-deterministic models.
- This `random_seed` param gets passed on to the `cross_validate` function ensuring that when it's set, the `set_random_seed` function is called right before setting each model.
- Since our LGBM and Random Forest models also rely on passing the `random_seed` into the wrapped model, the `cross_validate` function also does this whenever a model has a `random_seed` argument
- To test this works as expected, see the update to `TestPyTorchBackend`
- Also updated the tests for each of the models we have:
    - GP & SVM: Test that setting random seed has no effect, as expected
    - CNP: *Ignore since we are deprecating*
    - RF: Test that setting random seed has the intended effect
    - LGBM: I tried to do the same as RF, but couldn't coax it into making different predictions, even with different hyperparameters

### Reviewers

Take a look at all the above changes and comment


tests/experimental/test_experimental_gaussian_process_exact.py
```diff
@@ -75,29 +75,36 @@ def test_device(sample_data_y2d, new_data_y2d, device):
...
+    model3.fit(x, y)
+    pred3 = model3.predict(x2)
+
     assert isinstance(pred1, DistributionLike)
     assert isinstance(pred2, DistributionLike)
+    assert isinstance(pred3, DistributionLike)
     assert torch.allclose(pred1.mean, pred2.mean)
     assert torch.allclose(pred1.variance, pred2.variance)
+
+    # TODO: These should be different, but do we care that they aren't?
```
sgreenbury, line 109: It might be the case that:
- we expect these to be close for different fits of the GP (perhaps it would be worth also testing the hyperparameters of the fitted model for similarity?)
- the fitting is not stochastic for this model (and others in this PR too) and has no dependence on seed
 
I think it might be worth documenting for each model in the init docstring the expected effect of providing a random seed.

autoemulate/experimental/data/utils.py
```diff
@@ -190,6 +192,28 @@ def _denormalize(
         return (x * x_std) + x_mean
 
 
+class RandomMixin:
+    """Mixin class for ensuring reproducibility in AutoEmulate models."""
+
+    def set_random_seed(self, seed: int = 42, deterministic: bool = True):
```
sgreenbury, line 198: Perhaps it makes sense for this to remain a `@staticmethod` since `self` is not referred to in the method?

tests/experimental/test_experimental_base.py
```diff
@@ -17,10 +18,8 @@ class DummyModel(PyTorchBackend):
         A dummy implementation of PyTorchBackend for testing purposes.
         """
 
-        def __init__(self, x=None, y=None, random_seed=None, **kwargs):
+        def __init__(self, x=None, y=None, **kwargs):
             super().__init__()
-            if random_seed is not None:
-                self.set_random_seed(random_seed)
             _, _ = x, y  # unused variables
             self.linear = nn.Linear(1, 1)
```
sgreenbury, line 24: Perhaps it would be worth retaining this for subclass API so that there remains an option (though not the default) to set the random seed as part of the init?

autoemulate/experimental/compare.py
```diff
@@ -21,7 +21,9 @@ def __init__(
         y: InputLike,
         models: list[type[Emulator]] | None = None,
         device: DeviceLike | None = None,
+        random_seed: int | None = None,
     ):
+        self.random_seed = random_seed
```
sgreenbury, line 26: I think we might need to set the random seed within this init since there is a random_split call on L42?

autoemulate/experimental/compare.py
```diff
@@ -84,7 +86,11 @@ def compare(
             best_score_idx = scores.index(max(scores))
             best_model_config = configs[best_score_idx]
             cv_results = cross_validate(
-                cv(), self.train_val.dataset, model_cls, **best_model_config
+                cv=cv(),
+                dataset=self.train_val.dataset,
+                model=model_cls,
+                random_seed=self.random_seed,
```
sgreenbury, line 92: I'm wondering if the random seed is set in the `AutoEmulate `__init__()` (https://github.com/alan-turing-institute/autoemulate/pull/512#discussion_r2164555547) it can instead be passed here as `None`? Or perhaps it could be passed as a separate variable (e.g. `cv_seed`)?

edwardchalstrey1: Not sure exactly what you mean here? If we passed it to the default CV model which is KFold that expects random_state so we could do:
`cv=cv(random_state=self.random_seed)`

sgreenbury: So yeah to add to this, I'm wondering if we should  distinguish between:
1. setting a random seed at the top-level `AutoEmulate` (in the init)
2. setting the random seed at the cross-validation level within the loop over models
for the `AutoEmulate` class?

I think for the compare loop it might make sense as default to set once and then pass `None` (or a variable like `cv_seed: int | None = None`) to the cross-validate call. This should be deterministic still at the `AutoEmulate` level providing all inner loops are deterministic.

But an option to make this not `None` and be able to set seed for the cross-validate call might be useful too (e.g. to fix model seed at this point and/or the cv splits too as you mention too in case it is useful to control the randomness at that level).

tests/experimental/test_experimental_base.py
edwardchalstrey1: Did you mean for all Emulators or a particular subclass? The line you've commented here is in the dummy implementation of PyTorchBackend

sgreenbury: I think it makes sense to be part of a subclass specific model API (but with default typically `seed = None`) since not all models have any randomness (the dummy impl is the only example at the moment but more to be added soon and this would be an example impl) - what do you think?

edwardchalstrey1: Yeah I'm with you, have updated that, so for the models we have:
- LGBM and RF have the random_seed arg, but is passed into underlying model so this line isn't needed
- This dummy example (and future implemented models) we have this in the model init:
    ```
        if random_seed is not None:
                set_random_seed(random_seed)
    ```
- Other models do not have random_seed arg

autoemulate/experimental/compare.py
edwardchalstrey1: Not sure I understand the reason for this? At the moment the only reason to set the random seed in the AutoEmulate init is for the specific purpose of passing it through to our cross_validate function in model_selection such that we call `set_random_seed(seed=random_seed)` right before fitting the model (and also pass the arg directly to RF/LGBM)

edwardchalstrey1:
Note: we don't need to set seed before each model fit in the compare loop

sgreenbury (approved):
This looks great, thanks @edwardchalstrey1!

Issue #492, opened 2025-05-29
Simulator code contribution tutorial


edwardchalstrey1:
After v1.0 release and #469 complete, we should add a tutorial that explains to an AutoEmulate user how to set up a Simulator, similar to the `NaghaviSimulator` set up by @marjanfamili 

The tutorial may need to explain to a user how to:
- Create the class for the simulator
- Contribute install instructions
- Add a test for their simulator

EdwinB12:
@edwardchalstrey1 - Do you think we could do this with the Naghavi Simulator? 

I ask, because at the moment, Naghavi Simulator is in the codebase as an implemented Simulator. However, that means to test it, we need to depend on `    from ModularCirc.Models.NaghaviModel import NaghaviModel, NaghaviModelParameters`. I don't specific simulators should be a dependency for autoemulate.

edwardchalstrey1:
> [@edwardchalstrey1](https://github.com/edwardchalstrey1) - Do you think we could do this with the Naghavi Simulator?
> 
> I ask, because at the moment, Naghavi Simulator is in the codebase as an implemented Simulator. However, that means to test it, we need to depend on ` from ModularCirc.Models.NaghaviModel import NaghaviModel, NaghaviModelParameters`. I don't specific simulators should be a dependency for autoemulate.

@EdwinB12 Interesting point, do you think that Naghavi Simulator and all further simulators should be set up as external repos following a template? This might look similar to what we did on the [scivision project](https://scivision.readthedocs.io/en/latest/model_repository_template.html), but called something like "Integrate your simulator in the AutoEmulate active learning pipeline..." or something catchier

EdwinB12:
Intruiging. A possibly good thing to do in the long term. I think in the short term, we can simply build it in the notebook.

EdwinB12:
Yup, I'm going to do it in #485 but I'll make an issue.

PR #520, opened 2025-06-11
Update Simulator base class

### LINKED ISSUE #414: Refactor: combine simulator implementations
Currently there are two base class implementations for a simulator that had different purposes (active learning and application to cardiac modelling).

As part of the refactor, we should aim to harmonise the two implementations and ensure a subsequent refactored base class:
- has all required functionality for current applications
- has any envisaged required implementations for future applications

radka-j:
Closes #414

## Overview
This PR:
- primarily just moves new base Simulator from #485 with associated tests and updates to learners
- removes the `output_variables` argument as input to `Simulator` as it was specific to the `NaghaviSimulator` implementation rather than a generic simulator argument
- updates ActiveLearning notebook to use this Simulator
- adds a Projectile and Epidemic simulators

I left updates to `NaghaviSimulator `to be tackled as a separate PR (see discussion in #492). 

## Questions
- When creating this PR I run into an issue when the `sample_inputs` method started retuning tensors of float64 whereas everywhere else in the code base the tensors are float32 (this is the torch default). I have no idea why this happened and it made me think we might want to enforce consistency of types across the codebase. This could be us picking a type to stick to or enforcing a type based on inputs. Should we open an issue for that?


tests/experimental/test_experimental_base_simulator.py
```diff
@@ -0,0 +1,198 @@
...
+    # Should raise TypeError when trying to instantiate abstract base class
+    with pytest.raises(TypeError):
+        Simulator(params, outputs)  # type: ignore[abstract]
+
+
+def test_handle_simulation_failure():
+    """Test handling of simulation failures in forward_batch"""
+
+    # This test demonstrates that the current implementation doesn't properly handle
+    # None returns (which would be a good enhancement)
```
sgreenbury, line 168: Looking at this I was unsure where the `None` handling is demonstrated / tested in this test?

radka-j: It's more there as a reminder (not a test) that at the moment we don't handle the simulator failing. The NaghaviSimulator currently returns None but in this set up we don't allow for that. It's something to be addressed as part of #438

tests/experimental/test_experimental_base_simulator.py
```diff
@@ -0,0 +1,198 @@
...
+    class ThresholdSimulator(MockSimulator):
+        def _forward(self, x: TensorLike) -> TensorLike:
+            # Only process inputs where the first value is > 0.5
+            if x[0, 0] > 0.5:
+                return super()._forward(x)
+            # For other cases, let's return a tensor of zeros instead of None
+            # since the type system doesn't allow None returns
+            return torch.zeros((x.shape[0], len(self._output_names)))
```
sgreenbury, line 176: Perhaps we can add:
```suggestion
            # For other cases, let's return a tensor of zeros instead of None
            # since the type system doesn't allow None returns
            # TODO (#438): update to handle failed simulations
            return torch.zeros((x.shape[0], len(self._output_names)))
```

autoemulate/experimental/simulations/base.py
sgreenbury, line 1: I think the new cell outputs might need to be included here given the current way the docs are published without rerunning with jupyterbook

sgreenbury (approved):
I think this looks really great, thanks @radka-j @EdwinB12! I left a couple of small comments in the code but otherwise think this looks good to merge.

docs/tutorials/04_active_learning.ipynb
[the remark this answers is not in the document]
radka-j: I removed them because when I run the notebook and committed the output I got a git error complaining about the commit size.

PR #539, opened 2025-06-17
Refactor sensitivity analysis

### LINKED ISSUE #468: Refactor: sensitivity analysis
Functionality currently supported in main but not in experimental.

radka-j:
Closes #468 

In current AE version, sensitivity analysis is added as an additional method to the AutoEmulate object. Here I make it a separate object that the user can use independently.

### DIFF
--- a/autoemulate/experimental/sensitivity_analysis.py
+++ b/autoemulate/experimental/sensitivity_analysis.py
@@ -56,7 +56,6 @@ class SensitivityAnalysis(ConversionMixin):
         if problem is not None:
             problem = self._check_problem(problem)
         elif x is not None:
-            self.x, _ = self._convert_to_numpy(x)
             problem = self._generate_problem(x)
         else:
             msg = "Either problem or x must be provided."


autoemulate/experimental/sensitivity_analysis.py
```diff
@@ -0,0 +1,264 @@
...
+from SALib.sample.morris import sample as morris_sample
+from SALib.sample.sobol import sample as sobol_sample
+
+from autoemulate.experimental.data.utils import ConversionMixin
+from autoemulate.experimental.emulators.base import Emulator
+from autoemulate.experimental.types import DistributionLike, NumpyLike, TensorLike
+
+# NOTE: we still use these functions from main
+# should we just move them to experimental as well?
+from autoemulate.sensitivity_analysis import (
```
sgreenbury, line 15: This is a good point as I don't think we have strategy yet for combining experimental and current un-refactored code - perhaps we should have an issue for it and approach this as a last step before 1.0 importing where we can for now?

radka-j: Yes having an issue for this would be useful.

autoemulate/experimental/sensitivity_analysis.py
```diff
@@ -0,0 +1,264 @@
...
+                Si = morris_analyze(self.problem, param_samples, y[:, i])
+            results[name] = Si  # type: ignore PGH003
+
+        if method == "sobol":
+            return _sobol_results_to_df(results)
+        if method == "morris":
+            return _morris_results_to_df(results, self.problem)
+
+        # this should never happen, just keeping type checker happy
+        return None
```
sgreenbury, line 222: I think given the check on L200 we might be able to refactor this to (and remove the `None` from the return type):
```suggestion
        return _morris_results_to_df(results, self.problem)
```

autoemulate/experimental/sensitivity_analysis.py
```diff
@@ -0,0 +1,264 @@
...
+            raise ValueError(msg)
+
+        return output_names
+
+    def run(
+        self,
+        method: str = "sobol",
+        N: int = 1024,
+        conf_level: float = 0.95,
+    ) -> pd.DataFrame | None:
```
sgreenbury, line 169: Related to https://github.com/alan-turing-institute/autoemulate/pull/539/files#r2154672187
```suggestion
    ) -> pd.DataFrame:
```

autoemulate/experimental/sensitivity_analysis.py
radka-j: See #544

sgreenbury (approved):
Looks great, thanks @radka-j!

PR #561, opened 2025-06-23
Add more emulator subclasses

### LINKED ISSUE #542: Add EmulatorWithUncertainty subclass
from @sgreenbury in #504 
> subclass of Emulator (e.g. EmulatorWithUncertainty) that always predicts with DistributionLike (not just OutputLike) as the return type from predict and then we could update the type of the emulator here (and elsewhere) to be that subclass. We might also consider a subclass that predicts with GaussianLike too.

sgreenbury:
Closes #542.

This PR:
- Adds three new subclasses: `DeterministicEmulator`, `ProbabilisiticEmulator`, `GaussianEmulator`
- Updates current experimental emulators with new subclasses
- Does not affect any functionality since the subclasses currently defer predict calls to `super()`


edwardchalstrey1:
Might be worth drawing up an updated inheritance tree diagram for our Emulators, and also thinking about what level(s) we should be applying testing. I guess here you are doing the basic output type assertions in the classes themselves but would we want separate tests for each of these  - at the moment in `test_experimental_base.py` we just have `TestPyTorchBackend`

radka-j:
As suggested in #504, it would be great for the probabilistic emulators to have a method for returning mean and variance of a prediction as currently done in the `HistoryMatching.emulator_predict()` method. Once implemented, the HM method can be removed as well as removing passing an optional `Emulator` to the class at instantiation.

radka-j (approved):
This looks great!

radka-j:
Handling that in a separate PR sounds like a good way forward.

sgreenbury:
I've updated this now with:
- Revised the type hints where applicable given the change to `GaussianLike`
- Raised exceptions when instances are note `GaussianProcessLike` during transforms referring to #579
- Added initial implementations for the transforms to handle the new `GaussianLike` case though this code should be currently unreachable due to the errors being raised above

autoemulate/experimental/transforms/base.py
```diff
@@ -170,16 +195,138 @@ def _inverse_gaussian(self, y: GaussianLike) -> GaussianLike:
...
+        raise TypeError(msg)
+
+
+def _inverse_sample_gaussian_like(
+    c: Callable,
+    y: DistributionLike,
+    n_samples: int = 1000,
+    full_covariance: bool = True,
+) -> GaussianLike:
+    """Transforms a `DistributionLike`to a `GaussianLike` through sampling from `y`.
```
radka-j, line 238:
```suggestion
    """Transforms a `DistributionLike` to a `GaussianLike` through sampling from `y`.
```

autoemulate/experimental/transforms/base.py
```diff
@@ -170,16 +195,138 @@ def _inverse_gaussian(self, y: GaussianLike) -> GaussianLike:
...
+
+    Parameters
+    ----------
+    c : Callable
+        A callable that applies a transformation to generated the samples.
```
radka-j, line 243:
```suggestion
        A callable that applies a transformation to the generated samples.
```

autoemulate/experimental/transforms/base.py
```diff
@@ -170,16 +195,138 @@ def _inverse_gaussian(self, y: GaussianLike) -> GaussianLike:
...
+    n_samples: int = 1000,
+    full_covariance: bool = True,
+) -> GaussianProcessLike:
+    """Transforms a `GaussianProcessLike` to another `GaussianProcessLike` through
+    sampling from `y`.
+
+    Parameters
+    ----------
+    c : Callable
+        A callable that applies a transformation to generated the samples.
```
radka-j, line 298:
```suggestion
        A callable that applies a transformation to the generated samples.
```

radka-j (approved):
LGTM (just minor docstring comments)

autoemulate/experimental/emulators/base.py
```diff
@@ -92,6 +100,75 @@ def __init__(lr, batch_size):
...
+    @abstractmethod
+    def _predict(self, x: TensorLike) -> GaussianLike: ...
+    def predict(self, x: TensorLike) -> GaussianLike:
+        pred = super().predict(x)
+        assert isinstance(pred, GaussianLike)
+        return pred
+
+
+class GaussianProcessEmulator(GaussianEmulator):
+    """An emulator subclass that predicts with Gaussian outputs returning a
```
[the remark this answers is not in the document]
sgreenbury: Updated in e14f135b005f94e4fbfcaafaf68cec91731a09fa

PR #568, opened 2025-06-25
Add Polynomials to experimental

### LINKED ISSUE #547: Refactor: port second order polynomials

edwardchalstrey1:
Closes #547


radka-j:
I think for this one we should investigate what an implementation in PyTorch would look like and whether there is a performance loss to reimplementing it that way. I think it should be straightforward to do and I imagine that even doing SGD instead of an analytical solution (I think that's what sklearn does) shouldn't be too slow in this case given it's a simple regression model.

edwardchalstrey1:
> I think for this one we should investigate what an implementation in PyTorch would look like and whether there is a performance loss to reimplementing it that way. I think it should be straightforward to do and I imagine that even doing SGD instead of an analytical solution shouldn't be too slow given it's a simple regression model.

@radka-j I have refactored to be based on `PyTorchBackend` what do you think of this implementation? Still uses `from sklearn.preprocessing import PolynomialFeatures` but the model is now `nn.linear` - does involve some conversions back and forth but I think that's ok? Also adds some tuning hyperparameters as this was blank before

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -0,0 +1,67 @@
...
+        self.batch_size = batch_size
+
+        # Fit the polynomial feature transformer on the input data
+        self.poly = PolynomialFeatures(degree=self.degree)
+        x_np, _ = self._convert_to_numpy(x)
+        x_poly = self.poly.fit_transform(x_np)
+        x_poly_tensor = torch.tensor(x_poly, dtype=torch.float32, device=self.device)
+        self.n_features_in_ = x.shape[1]
+        self.n_outputs_ = y.shape[1] if y.ndim > 1 else 1
+        self.linear = nn.Linear(x_poly_tensor.shape[1], self.n_outputs_, bias=True).to(
```
radka-j, line 45:
```suggestion
        self.linear = nn.Linear(x_poly_tensor.shape[1], self.n_outputs_, bias=False).to(
```

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -0,0 +1,67 @@
...
+        self.lr = lr
+        self.epochs = epochs
+        self.batch_size = batch_size
+
+        # Fit the polynomial feature transformer on the input data
+        self.poly = PolynomialFeatures(degree=self.degree)
+        x_np, _ = self._convert_to_numpy(x)
+        x_poly = self.poly.fit_transform(x_np)
+        x_poly_tensor = torch.tensor(x_poly, dtype=torch.float32, device=self.device)
+        self.n_features_in_ = x.shape[1]
```
radka-j, line 43: do we need this?
```suggestion
```

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -0,0 +1,67 @@
...
+        self.degree = degree
+        self.lr = lr
+        self.epochs = epochs
+        self.batch_size = batch_size
+
+        # Fit the polynomial feature transformer on the input data
+        self.poly = PolynomialFeatures(degree=self.degree)
+        x_np, _ = self._convert_to_numpy(x)
+        x_poly = self.poly.fit_transform(x_np)
+        x_poly_tensor = torch.tensor(x_poly, dtype=torch.float32, device=self.device)
```
radka-j, line 42:
```suggestion

```

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -0,0 +1,67 @@
...
+        self.n_features_in_ = x.shape[1]
+        self.n_outputs_ = y.shape[1] if y.ndim > 1 else 1
+        self.linear = nn.Linear(x_poly_tensor.shape[1], self.n_outputs_, bias=False).to(
```
radka-j, line 44:
```suggestion
        self.linear = nn.Linear(x_poly.shape[1], self.n_outputs_, bias=False).to(
```

edwardchalstrey1:
- [ ] Test against the sklearn implementation

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -0,0 +1,66 @@
...
+        return self.linear(x_poly_tensor)
+
+    @staticmethod
+    def is_multioutput() -> bool:
+        return True
+
+    @staticmethod
+    def get_tune_config():
+        return {
+            "lr": [1e-3, 1e-2, 1e-1],
```
sgreenbury, line 63:
```suggestion
            "lr": [1e-3, 1e-2, 1e-1, 2e-1],
```

PR #569, opened 2025-06-25
Execute nb when building docs except workflow example

### LINKED ISSUE #538: Revisit CI to generate notebook outputs for docs
Currently tutorials in docs display outputs directly. However, this results in large outputs being committed which will accumulate over time as identified in https://github.com/alan-turing-institute/autoemulate/pull/520#discussion_r2150538571.

This issue aims to re-enable the tutorials being built with jupyterbook in CI to publish gh-pages branch where there is just the latest docs as a single commit.

A PR might need to revisit https://github.com/alan-turing-institute/autoemulate/pull/446 where notebooks that take a long time to run caused an issue with building the docs.

radka-j:
Closes #538 

The reason we turned execution off was because the workflow example was getting interrupted. This should just skip the workflow execution but run everything else. It feels we should merge this before merging #554

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -9,7 +9,9 @@ theme:
 # Force re-execution of notebooks on each build.
 # See https://jupyterbook.org/content/execute.html
 execute:
-  execute_notebooks: 'off'
+  execute_notebooks: 'force'
+  exclude_patterns:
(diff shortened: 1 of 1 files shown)


sgreenbury (approved):
LGTM!

PR #573, opened 2025-06-26
Docs restructure

### LINKED ISSUE #529: Restructure docs
It would be good to have a separate docs page for each of the downstream tasks:
- sensitivity analysis
- history matching
- calibration 

All should be within the same section in tutorials (e.g., `Downstream tasks`). Currently sensitivity analysis is within the "Deeper dive" notebook and the rest is within the "Workflow" notebook. They should use a projectile or epidemic as the simulation example.  This allows users to use each of those on their own rather than in an end to end way.

We should then have a separate notebook just for adding a simulator (#492) and then move the Naghavi workflow to a `case studies/` directory.

### LINKED ISSUE #572: Tutorials - keyboard interrupt issues
We re-enabled execution of notebooks when building documentation to avoid having to push large diffs to GitHub when we are required to run notebooks locally for the output to show.

Originally we disabled this because the workflow tutorial was taking too long and we encountered a `keyboardinterrupt` error in the outputs. We now skip this notebook but see this error in a number of notebooks
- scaling sensitivity
- active learning
- diffusion example

radka-j:
Closes #529
Closes #572

I increased the notebook execution time when building docs to make sure they all build but also chose to cache the results so that we don't have to re-execute everything each time as this would get very lengthy. This way the notebooks only get rerun when the code cells change. This won't catch if something breaks due to changes in the codebase.

The pages built fine when I run the build locally.


docs/experimental/index.md
```diff
@@ -0,0 +1,3 @@
+# Experimental
+
+This section documents use of experimental features that are under active development but not part of the main package yet.
```
edwardchalstrey1, line 3: Would it be better to just have the active learning and any other experimental notebooks in the same place but suppress publishing them to the site, or do we want people to read in-progress work online?

autoemulate/case_studies/blood_pressure/README.md
sgreenbury, line 1: I think the `case_studies/` path might be better up one directory level so that it is fully separate to the AE package?

PR #575, opened 2025-06-26
Refactor HMC calibration

### LINKED ISSUE #521: Refactor: MCMC calibration
In addition to the PyTorch refactor and integrating with the rest of the `experimental/` code, some things to consider:
- allow the user to pass multiple observations and handle correctly
- allow the user to run without passing sensitivity results
- possibly modularise by moving all the functionality that refines parameter bounds given NROY samples into a different class (using history matching for this is a very specific workflow that not everyone uses)
- we could also look into other plotting libraries (e.g., [arviz](https://python.arviz.org/en/stable/))

radka-j:
Closes #521 

Also adds new methods to `SensitivityAnalysis` and `HistoryMatching` to enable the full workflow currently implemented in main. 

I renamed it to `BayesianCalibration` as that is the general term for what we're doing. It has a `run_mcmc` method. We could imagine one day having a `run_svi` method in the same class. Not that it's a priority, just explaining motivation for the naming.

Start with the `hm_refactor.ipynb` notebook.


autoemulate/experimental/calibration/hmc.py
```diff
@@ -0,0 +1,238 @@
...
+            msg = "Noise must be either a float or a dictionary of floats."
+            raise ValueError(msg)
+
+    def model(self, predict: bool = False):
+        """
+        Pyro model.
+
+        Parameters
+        ----------
+        predict: bool
```
radka-j, line 102: The description here isn't correct. When the model is run with `predict=True` we aren't actually sampling. The posterior predictive is returned from the generated parameter MCMC samples.

radka-j:
I've also re-run the MCMC with the same data and emulator and there's variability in time there too (e.g., from 20-60 seconds).

autoemulate/experimental/calibration/bayes.py
```diff
@@ -0,0 +1,290 @@
...
+        Parameters
+        ----------
+        mcmc: MCMC
+            The MCMC object.
+        posterior_predictive: bool
+            Whether to include posterior predictive samples. Defaults to False.
+
+        Returns
+        -------
+        az.data.inference_data.InferenceData
```
sgreenbury, line 264:
```suggestion
        az.InferenceData
```

sgreenbury (approved):
This looks really great! Thanks @radka-j!

PR #581, opened 2025-06-30
Fix MLP initialization

### LINKED ISSUE #580: Fix epochs argument missing from MLP
The MLP doesn't enable the epochs to be set from the `__init__` API.

sgreenbury:
Closes #580.

This PR adds `epochs` and `bias_init` to the `MLP` constructor so that they are set as expected upon construction.


cisprague (approved):
Looks good – thanks!

PR #582, opened 2025-06-30
Add learning rate scheduler

### LINKED ISSUE #576: Add learning rate scheduler as hyperparameter
### Summary

Update PytorchBackend and GPytorchBackend to include a learning rate scheduler, see [PyTorch docs](https://docs.pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate)

- This would be an additional model config parameter
- May want to also include other params e.g. gamma
    - Things like which scheduler, optimizer to use (see momentum param below used by SGD but not adam)
- Note: Look at the MLP emulator in experimental where these params are already passed
- **Look into passing in the scheduler as an argument, instead of the individual hyperparameters** (partially instantiated e.g. with a default learning rate set)

### Basic Example

PyTorch example:
```
optimizer = optim.SGD(model.parameters(), lr=0.01, momentum=0.9)
scheduler = ExponentialLR(optimizer, gamma=0.9)

for epoch in range(20):
    for input, target in dataset:
        optimizer.zero_grad()
        output = model(input)
        loss = loss_fn(output, target)
        loss.backward()
        optimizer.step()
    scheduler.step()
```

### TODO:

- [x] Add optimizer as an argument to `PyTorchBackend` with a default
- [x] Add learning rate as an argument to `PyTorchBackend` with a default
- [x] Add scheduler argument to `PyTorchBackend`
- [x] Add scheduler in training loop
- [x] Add gamma as an argument to `PyTorchBackend` with a default

### Drawbacks

None, this would be done as part of 1.0 release instead of #475 

### Unresolved questions

_No response_

### Implementation PR

_No response_

### Reference Issues

_No response_

edwardchalstrey1:
Closes #576 

### Questions for reviewers

1. Are we happy with the [ExponentialLR](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ExponentialLR.html) or do we want to have a [set step size ](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.StepLR.html#steplr) **or** in the AutoEmulate context do we want to try both i.e. a randomly selected hyperparamater could be which scheduler to use
2. Similarly are we happy with using [Adam](https://docs.pytorch.org/docs/stable/generated/torch.optim.Adam.html#adam) optimizer or is this something worth varying?


autoemulate/experimental/emulators/base.py
```diff
@@ -107,7 +108,10 @@ class PyTorchBackend(nn.Module, Emulator, Preprocessor):
     verbose: bool = False
     preprocessor: Preprocessor | None = None
     loss_fn: nn.Module = nn.MSELoss()
+    optimizer_cls: type[optim.Optimizer] = optim.Adam
     optimizer: optim.Optimizer
+    lr: float = 1e-1
+    scheduler: ExponentialLR | None = None
```
edwardchalstrey1, line 114: Maybe we don't need a `None` option here, but I guess it gives the option to not use a scheduler if that was wanted

edwardchalstrey1: Can we do a base scheduler class here?

tests/experimental/test_experimental_base.py
```diff
@@ -27,7 +28,8 @@ def __init__(self, x=None, y=None, random_seed=None, device=None, **kwargs):
             _, _ = x, y  # unused variables
             self.linear = nn.Linear(1, 1)
             self.loss_func = nn.MSELoss()
-            self.optimizer = optim.SGD(self.parameters(), lr=0.01)
+            self.optimizer = self.optimizer_cls(self.parameters(), lr=self.lr)  # type: ignore[call-arg]
+            self.scheduler = ExponentialLR(self.optimizer, gamma=self.gamma)
```
radka-j, line 32: Same comment as for MLP
```suggestion

```

autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -46,6 +47,11 @@ class GaussianProcessExact(
 
     """
 
+    # TODO: refactor to work more like PyTorchBackend once any subclasses implemented
```
edwardchalstrey1, line 50: Since this is supposed to be a base class but also the only implementation we have so far I have explicitly defined the optmizer and scheduler in the `__init__` below. Does this make sense and should I set the scheduler to `None`?

edwardchalstrey1:
> Set up an __init__ in PyTorchBackend that initialises scheduler, optimizer and any other common params and call super init in the inits of the emulator subclasses of PyTorchBackend
Optimizer is different in MLP vs Polynomial
> Support for LRScheduler type hint could have a union scheduler_cls: type[LRScheduler] | partial[LRScheduler] where if the type is passed then it picks default values for any scheduler hyperparams
#591 
Have opened a new issue to look at post 1.0 release after discussion with @sgreenbury see

autoemulate/experimental/emulators/base.py
```diff
@@ -190,7 +192,12 @@ class PyTorchBackend(nn.Module, Emulator, Preprocessor):
     verbose: bool = False
     preprocessor: Preprocessor | None = None
     loss_fn: nn.Module = nn.MSELoss()
+    optimizer_cls: type[optim.Optimizer] = optim.Adam
     optimizer: optim.Optimizer
+    lr: float = 1e-1
+    scheduler_cls: type[LRScheduler] | partial[LRScheduler] = ExponentialLR
+    scheduler: LRScheduler | None = None
+    gamma: float = 0.9
```
radka-j, line 200: I think we could do something like this everywhere
```suggestion
        scheduler_cls: type[LRScheduler] | partial[LRScheduler] | None = None
```

autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -121,6 +129,11 @@ def __init__(  # noqa: PLR0913 allow too many arguments since all currently requ
         self.lr = lr
         self.batch_size = batch_size
         self.activation = activation
+        self.gamma = gamma
+        self.optimizer = self.optimizer_cls(self.parameters(), lr=self.lr)  # type: ignore[call-arg] since all optimizers include lr
+        self.scheduler = None
```
radka-j, line 134:
```suggestion
        self.optimizer = self.optimizer_cls(self.parameters(), lr=self.lr)  # type: ignore[call-arg] since all optimizers include lr
        if scheduler_cls is None:
                self.scheduler = None
        else:
                # here need to also handle whether it's partial or not and do the right thing 
```

edwardchalstrey1:
TODO:
- [ ] Add tests where scheduler is and isn't set so we can have it None by default

autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -121,6 +129,11 @@ def __init__(  # noqa: PLR0913 allow too many arguments since all currently requ
         self.lr = lr
         self.batch_size = batch_size
         self.activation = activation
+        self.gamma = gamma
```
radka-j, line 132:
```suggestion

```

edwardchalstrey1:
@radka-j @sgreenbury  see if you like the way I've implemented hyperparameter selection for the scheduler

radka-j:
> @radka-j @sgreenbury see if you like the way I've implemented hyperparameter selection for the scheduler

looks very nice!

edwardchalstrey1:
Now includes changes from #586 so merge that first

edwardchalstrey1:
@sgreenbury @cisprague do you understand why [this test](https://github.com/alan-turing-institute/autoemulate/actions/runs/16145931849/job/45564658727?pr=582) fails? Looks like it's related to #586

autoemulate/experimental/emulators/base.py
```diff
@@ -204,6 +209,56 @@ def loss_func(self, y_pred, y_true):
...
+        """
+        all_params = [
+            {
+                "scheduler_cls": [ExponentialLR],
+                "scheduler_kwargs": [
+                    {"gamma": 0.9},
+                    {"gamma": 0.95},
+                ],
+            },
+            # TODO: investigate these suggestions from copilot
```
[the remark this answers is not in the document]
sgreenbury: Perhaps we can add the one used in current `ExactGPRegressor` too or instead?
```suggestion
            {
                "scheduler_cls": LRScheduler,
                "scheduler_kwargs": [{"policy":"ReduceLROnPlateau", "patience": 5, "factor": 0.5)}]
            },
            # TODO: investigate these suggestions from copilot
```

autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -133,6 +140,13 @@ def __init__(  # noqa: PLR0913 allow too many arguments since all currently requ
         self.epochs = epochs
         self.lr = lr
         self.activation = activation
+        self.optimizer = self.optimizer_cls(self.parameters(), lr=self.lr)  # type: ignore[call-arg] since all optimizers include lr
+        # Extract scheduler-specific kwargs if present
+        scheduler_kwargs = kwargs.pop("scheduler_kwargs", {})
+        if self.scheduler_cls is None:
+            self.scheduler = None
+        else:
+            self.scheduler = self.scheduler_cls(self.optimizer, **scheduler_kwargs)
```
sgreenbury, line 149: Perhaps it would be worth refactoring this into a function or method on a base class (e.g. `construct_scheduler`) as repeated a few times? An alternative to having to init the other variables too like the optimizer etc would be good but I think that would make sense in a new issue/PR.

sgreenbury (commented):
I think this looks great, thanks @edwardchalstrey1! I left two small suggestions above and also a suggestion for the failing CI and then think it looks good to merge with #586 now merged.

cisprague:
The covariance matrices returned from`Ensemble` emulators should at least be symmetric and have non-negative eigenvalues (positive **semi**-definite). So an addition of some small jitter on the diagonals should work, and symmetrising with `(A + A.T)/2` wouldn't hurt (it should already be symmetric). Otherwise, it could just be some round-off error – some eigenvalues might be like `-1e-6` for large matrices.

sgreenbury (approved):
This looks good to merge, thanks @edwardchalstrey1!

PR #584, opened 2025-07-01
Add refit method and tests

### LINKED ISSUE #583: Implement refit method for experimental AutoEmulate object
Currently, there is no in-built way of getting the trained models out of the `compare` method. 

Add a `refit` method that does the following: 

- Takes in X and Y so model can be fit on any data 
- By default, trains the best model from the metrics calculated by `compare`
- Allows an argument where you pass the model name you would like to refit.

EdwinB12:
Closes #583

### DIFF
--- a/autoemulate/experimental/compare.py
+++ b/autoemulate/experimental/compare.py
@@ -142,10 +142,10 @@ class AutoEmulate(ConversionMixin, TorchDeviceMixin):
             Metric to use for selecting the best model. Default is 'r2_score'.
         x : InputLike | None, optional
             Input data to use for refitting. If None, uses the original data
-            passed to AutoEmulate constructor.
+            passed to AutoEmulate constructor. Must be provided together with y.
         y : InputLike | None, optional
             Target data to use for refitting. If None, uses the original data
-            passed to AutoEmulate constructor.
+            passed to AutoEmulate constructor. Must be provided together with x.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


autoemulate/experimental/compare.py
```diff
@@ -109,4 +113,104 @@ def compare(
                 "rmse_score": rmse_score,
             }
             self.log_compare(model_cls, best_model_config, r2_score, rmse_score)
+
+        # Store comparison results for refit method
+        self._comparison_results = models_evaluated
         return models_evaluated
+
+    def refit(
```
radka-j, line 121: At the moment the user can either refit on ALL the data (train+val+test) or on their custom dataset. Do we need an option to refit on just the train+val data. This is what AutoEmulate does at the moment as it allows the user to check the test performance of their model.

autoemulate/experimental/compare.py
```diff
@@ -109,4 +113,104 @@ def compare(
...
+        best_config = self._comparison_results[best_model_name]["config"]
+
+        # Use provided data or default to original data
+        if x is None and y is None:
+            refit_x, refit_y = self.x, self.y
+        elif x is not None and y is not None:
+            # Convert both x and y to tensors
+            refit_x, refit_y = self._convert_to_tensors(x, y)
+            refit_x, refit_y = self._move_tensors_to_device(refit_x, refit_y)
+        elif x is not None and y is None:
```
radka-j, line 198: I am not sure about the option of passing just the Xs or just the Ys - what is the use case for this?

autoemulate/experimental/compare.py
radka-j: Following comments in #588, the refit() method should by default fit on ALL the data (train+val+test) or whatever new dataset the user wants to pass it. Note that there are some additional proposed changes to `compare()` within that issue, which mean that we should not hold onto state as is implemented in this PR but rather just return the fitted models.

PR #587, opened 2025-07-03
Add EarlyStopping for GPs

radka-j:


autoemulate/experimental/emulators/early_stopping.py
```diff
@@ -0,0 +1,141 @@
...
+            self.dynamic_threshold_ = self._calc_new_threshold(curr_score)
+            self.best_epoch_ = curr_epoch
+            if self.load_best:
+                self.best_model_weights_ = deepcopy(model.state_dict())
+        if self.misses_ == self.patience:
+            print(
+                "Stopping since train loss has not improved in the last "
+                f"{self.patience} epochs."
+            )
+            raise KeyboardInterrupt
```
sgreenbury, line 97: Perhaps a custom exception (or one that can have a msg) so the print becomes part of the exception?
```suggestion
            msg = (
                "Stopping since train loss has not improved in the last "
                f"{self.patience} epochs."
            )
            raise EarlyStoppingException(msg)
```

sgreenbury (approved):
I think this looks great, thanks for adding this @radka-j! My main comment is that I think this should apply to a validation loss by default - perhaps we can open this as a new issue to extending this? Otherwise I left a few small comments but looks good to merge!

autoemulate/experimental/emulators/gaussian_process/exact.py
```diff
@@ -170,6 +179,16 @@ def _fit(self, x: TensorLike, y: TensorLike):
             self.log_epoch(epoch, loss)
             optimizer.step()
 
+            if self.early_stopping is not None:
+                try:
+                    self.early_stopping.on_epoch_end(self, epoch, loss.item())
```
sgreenbury, line 184: Might be worth adding a todo here relating to using a validation loss instead?

Issue #469, opened 2025-05-14
Refactor: update tutorials


sgreenbury:
This issue captures the aim of being able to run all the docs tutorials with code that is in experimental for v1.0.

radka-j:
We could split the tutorials into the following sections.

MODELLING and PREDICTION
- [ ] quickstart 
    - show off compare loop
    - plot results
- [x] dimensionality reduction 
- [ ] adding your own model

DOWNSTREAM TASKS
- [ ] Sensitivity analysis
    - show both morris and sobol
- [ ] History matching
    - metric
    - workflow
- [ ] MCMC calibration

SIMULATOR IN THE LOOP
- [x] Adding a simulator
- [x] Active learning

radka-j:
I wonder if the first section should also touch on:
- model configuration (e.g., select from  ALL_EMULATORS, ALL_PYTORCH_EMULATORS,  a list of specific ones)
- uncertainty quantification (not all models support this)

I am not sure where the existing ensembles tutorial fits in.

Issue #467, opened 2025-05-14
Refactor: update docs


sgreenbury:
This issue is for capturing all changes to the non-tutorial documentation that's required when moving the updated implementation from experimental to the main package.

This includes (TODO: add other areas):
- User guides
- API docs (docstrings) - full numpy-style
- ...

sgreenbury:
Check:
- [ ] Locally test docs for experimental

For day before release day:
- [ ] Rerun notebooks (check latest APIs)
- [ ] Update landing page of the docs website

sgreenbury:
Closing with merged #663 #664 and now final changes in #671.

PR #598, opened 2025-07-07
Start experimental docs

radka-j:
Contributes to #467 and especially #469

The focus of this PR is on restructuring the tutorials into subsections as described in the issue. We can then start adding updated tutorials with the experimental features there.


edwardchalstrey1 (approved):
LGTM

PR #603, opened 2025-07-08
Evaluate on test data, update compare loop and add Results/plots

### LINKED ISSUE #408: Refactor: plotting
The original AutoEmulate class comes with plotting functions that we might want to reimplement (can also update in the process).

### LINKED ISSUE #531: Add refactored transforms to compare loop

### LINKED ISSUE #588: Refactor: compare should refit on train+val data and evaluate on test data
We want to update the `compare()` loop to something like the following (note this is more pseudocode than exact version):

It should look something like:
```python
        for model_cls in self.models:
            scores, configs = tuner.run(model_cls)
            best_score_idx = scores.index(max(scores))
            best_model_config = configs[best_score_idx]
            
            # CHANGES START HERE

            # - we refit on all the `train_val` data - we were missing this step compared to current AE
            # - this is what the refit() method in #584 was going to do by default
            m = model_class(self.train_val.dataset, device=self.device, **model_config)
            m.fit(elf.train_val.dataset)

            # - we then evaluate on test data - this returns the same results as AE v0 `evaluate()` method
            y_pred = m.predict(self.test_x)
            score = evaluate(self.test_y, y_pred, self.metric, self.device)

            # - add fitted model to what is returned so the user has access to it
            # - this removes the need to store state as is currently implemented in #584 with `self._comparison_results`
            # - we're also changing this to a list rather than dictionary - is this the best choice?
            models_evaluated.append({
                "model_type":  model_cls.__name__,
                "config": best_model_config,
                "model": m,
               "score": score,
            })
        return models_evaluated
```

edwardchalstrey1:
Closes #588 
Closes #408 
Closes #531
    - Includes #551 

Pre-requisite PRs:
- #586 
- #582

### Changes

- Evaluate models on test data
- Implement Result/Results classes for storing results of model runs on data
- Implement plotting functions
- Quickstart notebook using v1/experimental AutoEmulate code
- #551


autoemulate/experimental/plotting.py
```diff
@@ -0,0 +1,169 @@
...
+        # we don't show otherwise it will double plot
+        plt.close(fig)
+        return fig
+    # in terminal, show the plot
+    plt.close(fig)
+    plt.show()
+    return fig
+
+
+def plot_Xy(  # noqa: PLR0913
```
sgreenbury, line 35: I think we should stick with lower case naming following the refactor? So e.g. `plot_xy(`

edwardchalstrey1:
TODO:
- [ ] merge main into here
- [ ] de-capitalise Xy instances and all upper case X's and Y's (only in experimental)
- [ ] Set id as the index integer instead of as a string

sgreenbury (approved):
Looks great @edwardchalstrey1, thanks! And good to merge once the mentioned changes above are added.

PR #604, opened 2025-07-08
Fix training only first batch

### LINKED ISSUE #594: Possible bug: cross-validate fits on a first batch
It appears that cross-validate only fits on the first batch https://github.com/alan-turing-institute/autoemulate/blob/caeafc4d5ee886846086df475aeb0c5988b111cc/autoemulate/experimental/model_selection.py#L119-L122

sgreenbury:
Closes #594.

### DIFF
--- a/autoemulate/experimental/data/utils.py
+++ b/autoemulate/experimental/data/utils.py
@@ -17,8 +17,9 @@ class ConversionMixin:
     Mixin class to convert input data to pytorch Datasets and DataLoaders.
     """
 
+    @classmethod
     def _convert_to_dataset(
-        self,
+        cls,
         x: InputLike,
         y: InputLike | None = None,
     ) -> Dataset:
@@ -49,8 +50,9 @@ class ConversionMixin:
 
         return dataset
 
+    @classmethod
     def _convert_to_dataloader(
-        self,
+        cls,
         x: InputLike,
         y: InputLike | None = None,
         batch_size: int = 16,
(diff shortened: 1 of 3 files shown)


autoemulate/experimental/data/utils.py
```diff
@@ -66,25 +68,29 @@ def _convert_to_dataloader(
...
 
         # Handle Subset of TensorDataset
         if isinstance(dataset, Subset):
-            if isinstance(dataset.dataset, TensorDataset):
-                tensors = dataset.dataset.tensors
+            ds = dataset
+            while isinstance(ds, Subset):
+                ds = ds.dataset
+            if isinstance(ds, TensorDataset):
+                tensors = ds.tensors
```
sgreenbury, line 93: Added this as I think there were cases of nested `Subset` - does the indices look correct for this?

autoemulate/experimental/data/utils.py
```diff
@@ -17,8 +17,9 @@ class ConversionMixin:
     Mixin class to convert input data to pytorch Datasets and DataLoaders.
     """
 
+    @classmethod
```
sgreenbury, line 20: Updated to `@classmethod` so that they can be called independently of an instance.

sgreenbury:
Rebased to introduce fix for CI here from #586

autoemulate/experimental/data/utils.py
sgreenbury: Sorry missed the line in the quote: https://github.com/alan-turing-institute/autoemulate/pull/604/files#r2192232697

PR #605, opened 2025-07-08
Grad handling

### LINKED ISSUE #602: Enable emulator predictions with grad
Requirement for #521 

Currently emulators return predictions without gradient information. Often the user just wants the prediction so this felt appropriate, but for some downstream tasks (e.g., MCMC) that prediction feeds into a computation that needs to be differentiated. 

A solution would be to return predictions with gradient information and any downstream code handle it. Another option would be to include a `with_grad` option in `predict()` to enable both. Is there another option? Which is better?

radka-j:
Closes #602

TODO:
- [ ] transformed emulator
- [ ] ensembles

### DIFF
--- a/autoemulate/experimental/emulators/base.py
+++ b/autoemulate/experimental/emulators/base.py
@@ -7,10 +7,7 @@ from sklearn.base import BaseEstimator
 from torch import nn, optim
 
 from autoemulate.experimental.data.preprocessors import Preprocessor
-from autoemulate.experimental.data.utils import (
-    ConversionMixin,
-    ValidationMixin,
-)
+from autoemulate.experimental.data.utils import ConversionMixin, ValidationMixin
 from autoemulate.experimental.device import TorchDeviceMixin
 from autoemulate.experimental.types import (
     DistributionLike,
@@ -51,16 +48,16 @@ class Emulator(ABC, ValidationMixin, ConversionMixin, TorchDeviceMixin):
         return cls.__name__
 
     @abstractmethod
-    def _predict(self, x: TensorLike) -> OutputLike:
+    def _predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
         pass
 
-    def predict(self, x: TensorLike) -> OutputLike:
+    def predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
         if not self.is_fitted_:
             msg = "Model is not fitted yet. Call fit() before predict()."
             raise RuntimeError(msg)
         self._check(x, None)
         (x,) = self._move_tensors_to_device(x)
-        output = self._predict(x)
+        output = self._predict(x, with_grad)
         self._check_output(output)
         return output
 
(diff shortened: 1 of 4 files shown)


autoemulate/experimental/emulators/transformed/base.py
```diff
@@ -367,14 +364,24 @@ def _fit(self, x: TensorLike, y: TensorLike):
...
+                return self._inv_transform_y_tensor(y_t_pred)
+        else:
+            with torch.no_grad():
+                # Transform and invert transform for prediction in original data space
+                x_t = self._transform_x(x)
+                y_t_pred = self.model.predict(x_t, with_grad)
+
+                # If TensorLike, transform tensor back to original space
+                if isinstance(y_t_pred, TensorLike):
+                    return self._inv_transform_y_tensor(y_t_pred)
```
sgreenbury, line 384:
```suggestion
        with torch.set_grad_enabled(with_grad):
            # Transform and invert transform for prediction in original data space
            x_t = self._transform_x(x)
            y_t_pred = self.model.predict(x_t, with_grad)

            # If TensorLike, transform tensor back to original space
            if isinstance(y_t_pred, TensorLike):
                return self._inv_transform_y_tensor(y_t_pred)
```

autoemulate/experimental/emulators/base.py
```diff
@@ -51,16 +48,16 @@ def model_name(cls) -> str:
         return cls.__name__
 
     @abstractmethod
-    def _predict(self, x: TensorLike) -> OutputLike:
+    def _predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
```
sgreenbury, line 51: Given there is the option to set_grad_enabled, perhaps we do not need to add the the `_predict()` API? This might be helpful so that subclassing would not need to consider this by default.

sgreenbury (commented):
I think this is looking great, thanks @radka-j!

My main thought is whether we might be able to refactor to only pass `with_grad` in the `predict()` public API and wrap the call within the default implementation with a context manager, what do you think? E.g.:
```python
def predict(self, x, with_grad):
    # ...
    # We could possibly add a new method to check if instance supports gradients?
    if not self.supports_grad():
        raise ...
    
    with torch.set_grad_enabled(with_grad):
        return self._predict(x)
```

radka-j:
@sgreenbury I updated the code to use `with torch.set_grad_enabled(with_grad):`, it looks much better now, thanks!! I had to use it inside the `_predict()` method rather than at the top level `predict()` because of the ensemble and transformed emulators.

sgreenbury (approved):

autoemulate/experimental/emulators/transformed/base.py
```diff
@@ -137,6 +133,11 @@ def __init__(  # noqa: PLR0913
         self.n_samples = n_samples
         self.full_covariance = full_covariance and y.shape[1] <= max_targets
         TorchDeviceMixin.__init__(self, device=device)
+        self.supports_grad = (
+            self.model.supports_grad
+            and all(t.bijective for t in self.x_transforms)
+            and all(t.bijective for t in self.y_transforms)
```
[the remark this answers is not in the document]
radka-j: Will do

PR #608, opened 2025-07-10
428 improve logging

### LINKED ISSUE #428: Refactor: Improve logging throughout codebase
We should remove any print statements and replace with logging. We should also format the logging in a central place.

EdwinB12:
- Setup formatting for a logger for the AutoEmulate class
- Add support for setting a logging level to the logger for AutoEmulate 
- Add support for progress bar without conflicting logging
- Remove progress bar from tuner

TODO: 
- If agree with approach, go and add logging throughout codebase.


EdwinB12:
Okay, will do. 

Remember to: 

- Remove debugging code in the __main__ == __name__ in compare.py
- Agree on a default

EdwinB12:
@radka-j - This is ready for review. Note that I have only added logging statements in functions currently called by the compare method.

EdwinB12:
@radka-j - Not sure how to handle this merge conflict. If we keep the debug logging when fitting models, it kind of overwhelmes the logging. On the other hand, it could be handy. .

EdwinB12:
I've removed logging from gp. Ready for Review @radka-j

radka-j (approved):
This looks great, thanks!

PR #620, opened 2025-07-14
Add bootstrap and revise results

### LINKED ISSUE #617: Add bootstrap estimator for estimating uncertainty in evaluation
Currently performance is measured on a single a held-out test split as part of the AutoEmulate compare loop.

To estimate the uncertainty in this measure, a statistical bootstrap could be included as part of the model selection module.

sgreenbury:
Closes #617.

### DIFF
--- a/autoemulate/experimental/compare.py
+++ b/autoemulate/experimental/compare.py
@@ -11,7 +11,7 @@ from autoemulate.experimental.emulators import ALL_EMULATORS
 from autoemulate.experimental.emulators.base import Emulator
 from autoemulate.experimental.emulators.transformed.base import TransformedEmulator
 from autoemulate.experimental.logging_config import configure_logging
-from autoemulate.experimental.model_selection import evaluate
+from autoemulate.experimental.model_selection import bootstrap, evaluate
 from autoemulate.experimental.plotting import (
     calculate_subplot_layout,
     display_figure,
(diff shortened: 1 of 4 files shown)


autoemulate/experimental/model_selection.py
```diff
@@ -24,31 +25,37 @@
...
 
 
 def evaluate(
-    y_true: InputLike,
     y_pred: OutputLike,
-    metric: type[torchmetrics.Metric],
-    device: DeviceLike,
+    y_true: InputLike,
+    metric: type[torchmetrics.Metric]
+    | partial[torchmetrics.Metric] = torchmetrics.R2Score,
```
sgreenbury, line 58: This update and elsewhere in this module is since the `torchmetrics` API expects predictions first and target values second in postional args.

radka-j (approved):
This looks great! I only had one question about letting user set `n_bootstraps`. This could handle the option of just using the test data once.

autoemulate/experimental/compare.py
```diff
@@ -307,7 +326,7 @@ def plot(  # noqa: PLR0912, PLR0915
         if isinstance(y_pred, DistributionLike):
             y_variance = y_pred.variance
```
edwardchalstrey1, line 327: Should we update the plot function and the `plot_xy` called within passing in `r2_test_std` instead of `y_variance`? I set it up previously to calculate the std within that function:
```
if y_variance is not None:
        y_variance_sorted = y_variance[sort_idx]
        y_std = np.sqrt(y_variance_sorted)
```

sgreenbury: Yeah I think we can keep the `plot_xy` as is `y_std` as implemented. Though it might be interesting to add plot functionality comparing the `R^2` across models with error bars for the $R^2$ to visually compare the models too?

Issue #302, opened 2025-02-18
Add argument docstrings and type hints to `LatinHypercube.__init__`


cisprague:
The `LatinHypercube.__init__` method is missing explicit argument docstrings and type hints. Adding these would improve clarity when interacting with the class.

Issue #555, opened 2025-06-20
Refactor: docstring style and ruff config


sgreenbury:
Following discussion in https://github.com/alan-turing-institute/autoemulate/pull/474#discussion_r2157236380, it would be good to revisit docstring style. Currently [CONTRIBUTING.md](https://github.com/alan-turing-institute/autoemulate/blob/b4faf29cf5671559163a6e4aa4d45295f7bdd314/CONTRIBUTING.md) indicates [numpy-style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html).

As part of this issue we can also implement linting for this with config in [ruff](https://docs.astral.sh/ruff/settings/#lintpydocstyle).

PR #626, opened 2025-07-14
Update docstrings

radka-j:
Contributes to #302

### DIFF
--- a/autoemulate/experimental/compare.py
+++ b/autoemulate/experimental/compare.py
@@ -184,6 +184,10 @@ class AutoEmulate(ConversionMixin, TorchDeviceMixin, Results):
         self.logger.debug(msg)
 
     def compare(self):
+        """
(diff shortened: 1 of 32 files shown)


sgreenbury (approved):
LGTM, thanks @radka-j!

One very minor thought was whether we should aim for consistency with "variable: type" or "variable : type" in the docstrings? I had a look to see if Ruff is specific/formats this when the it is checking docstrings or if we could use https://pypi.org/project/docformatter/ for this but it doesn't look like either of these tools address this. Perhaps one to come back to for #555?

radka-j:
@sgreenbury thank you for pointing that out because it bothers me that it isn't consistent but I was trying to not be too pedantic about it 😅 personally I prefer "variable: type" but I could be convinced otherwise.

radka-j:
It looks like the Numpy example does "variable : type" though, we should do whatever they do

Issue #454, opened 2025-05-01
Refactor: enhance the testing of compare


sgreenbury:
From discussion with @EdwinB12, an area we'll look to revisit once we've implemented a fuller pipeline in experimental is more extensive testing of different combinations with the items.

E.g. parameterize over:
- Input processor
- Emulator
- Output preprocessor

sgreenbury:
### Script and parameters
[Script]() run with:
- Simulators: `Epidemic`, `Projectile`, `ProjectileMultioutput`
- `n_iter` for tuning: 10, 50, 100, 200
- `n_samples` from simulator: 20, 50, 100, 200, 500
- `n_splits` for : 5 (default for AE)

### Results

[benchmark_results.csv](https://github.com/user-attachments/files/21387213/benchmark_results.csv)

### Plots

[plot_benchmark.ipynb]()

### Errors


### Warnings
 

### Next steps

sgreenbury:
### Script and parameters
[Script]() run with:
- Simulators: `Epidemic`, `Projectile`, `ProjectileMultioutput`
- `n_iter` for tuning: 10, 50, 100, 200
- `n_samples` from simulator: 20, 50, 100, 200, 500
- `n_splits` for : 2, 5 (default for AE)

### Results

[benchmark_results.csv](https://github.com/user-attachments/files/21401246/benchmark_results.csv)

### Plots

[plot_benchmark.ipynb]()

### Errors
TODO

### Warnings
TODO

### Next steps
TODO

sgreenbury:
### Script and parameters
[Script](https://github.com/alan-turing-institute/autoemulate/blob/e2e3cf52cadb18a84f68ce85ec87aa7dcf019813/benchmarks/benchmark.py) run with:
- Simulators: `Epidemic`, `FlowProblem`, `Projectile`, `ProjectileMultioutput`
- `n_iter` for tuning: 10, 50, 100, 150, 200
- `n_samples` from simulator: 20, 50, 100, 200, 500
- `n_splits` for : 2, 5 (default for AE)

### Results

[benchmark_results.csv](https://github.com/user-attachments/files/21469387/benchmark_results.csv)

### Plots

[plot_benchmark.ipynb]()

### Errors
TODO

### Warnings
TODO

### Next steps
- With increasing `n_iter` there is still some decrease in performance

sgreenbury:
[...]

PR #627, opened 2025-07-15
Fix missing model config to pass to cv

sgreenbury:
Following https://github.com/alan-turing-institute/autoemulate/issues/454#issuecomment-3072519163, fixed missing random config to be passed to cv.

### DIFF
--- a/autoemulate/experimental/model_selection.py
+++ b/autoemulate/experimental/model_selection.py
@@ -84,11 +84,11 @@ def cross_validate(  # noqa: PLR0913
     cv: BaseCrossValidator,
     dataset: Dataset,
     model: type[Emulator],
+    model_config: ModelConfig,
     x_transforms: list[AutoEmulateTransform] | None = None,
     y_transforms: list[AutoEmulateTransform] | None = None,
     device: DeviceLike = "cpu",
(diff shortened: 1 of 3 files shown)


sgreenbury:
> Does this mean that we were running CV with just the default model option each time?   😱

Yeah that was my conclusion too! I think it's from a recent change (in #603 following updating the compare loop) when switching to `cross_validate` in the tuner, apologies!

> Should we replace `kwargs` in cross_validate with an explicit `model_config` and add to docstrings?

Thanks, agreed renaming `**kwargs` as `**model_config` and adding docs would be better. I'll make the change.

autoemulate/experimental/model_selection.py
```diff
@@ -84,11 +83,11 @@ def cross_validate(  # noqa: PLR0913
     cv: BaseCrossValidator,
     dataset: Dataset,
     model: type[Emulator],
+    model_config: ModelConfig,
```
radka-j, line 86: Just a suggestion to avoid having to pass it if the user just wants to use the default? I can also see an argument against this as opening up the potential for user error.

```suggestion
    model_config: ModelConfig = {},
```

PR #628, opened 2025-07-15
Add logger to BayesianCalibrator

### LINKED ISSUE #612: Add logging to BayesianCalibrator
Following logging introduced in #608

edwardchalstrey1:
Closes #612 

### Reviewers

Please check if the message themselves make sense or if others should be added! Note the distinction between "info", "error" and "debug" messages

### DIFF
--- a/autoemulate/experimental/calibration/bayes.py
+++ b/autoemulate/experimental/calibration/bayes.py
@@ -74,6 +74,9 @@ class BayesianCalibration(TorchDeviceMixin):
         self.emulator.device = self.device
         self.output_names = list(observations.keys())
         self.logger, self.progress_bar = get_configured_logger(log_level)
+        self.logger.info(
+            "Initializing BayesianCalibration with parameters: %s", calibration_params
+        )
 
         # Check observation tensors are 1D (convert if 0D)
         processed_observations = {}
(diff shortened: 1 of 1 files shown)


autoemulate/experimental/calibration/bayes.py
```diff
@@ -135,22 +167,33 @@ def model(self, predict: bool = False):
                 min_val, max_val = self.parameter_range[param]
                 sampled_val = pyro.sample(param, dist.Uniform(min_val, max_val))
                 param_list.append(sampled_val.to(self.device))
+                self.logger.debug(
+                    "Sampled parameter '%s' from Uniform(%s, %s)",
+                    param,
+                    min_val,
+                    max_val,
+                )
```
radka-j, line 175: One usually does at least 1000k samples so this would probably be just overwhelming
```suggestion
```

radka-j (commented):
This is looking great! I mostly suggest we remove logging from within the model section as this gets called each time we sample (1000+ times). I'd also suggest we add an info message when `self.calibration_params` gets set.

autoemulate/experimental/calibration/bayes.py
```diff
@@ -63,35 +73,49 @@ def __init__(  # noqa: PLR0913
         self.emulator = emulator
         self.emulator.device = self.device
         self.output_names = list(observations.keys())
+        self.logger, self.progress_bar = get_configured_logger(log_level)
+        self.logger.info(
+            "Initializing BayesianCalibration with parameters: %s", calibration_params
```
radka-j, line 78: In case `calibration_params is None`

```suggestion
            "Initializing BayesianCalibration with parameters: %s", self.calibration_params
```

radka-j (approved):

PR #629, opened 2025-07-15
Add saving and loading models to experimental

### LINKED ISSUE #524: Refactor: option to save/load models
TODO:
- [x] Move save.py to experimental and test
- [x] API should be `ae.save(best_result)` or `ae.save(best_result.id)` or `ae.save(model)`
    - [x] If we make it possible to pass the Result object rather than just the id, ensure this is also true for `ae.plot`
- [x] also add ae.load
- [x] Update tests for compare to include save and load
- [x] Update the quickstart notebook with model saving and loading
    - [x] Commit this
- [x] Add option to save the rest of the result metadata?
- [x] Remove short_name feature of Emulator class since it is unused

### LINKED ISSUE #622: AE quickstart tutorial should demo key API functionality and have clear narrative
The quickstart notebook shows how to use AE as is. It'd be good to have the second notebook give an overview of  some of the customisation options, e.g.:
- choosing subsets of models
- changing the model training configs 
- learning rate scheduler options 
- logging
-  ...?

This is also a chance to go a bit more in depth into what is happening under the hood.

edwardchalstrey1:
Closes #524 
Closes #622 

### Changes

See checklist in #524 and discussion in #622


autoemulate/experimental/save.py
```diff
@@ -0,0 +1,155 @@
...
+            self.logger.info("%s saved to %s", model_name, full_path)
+            return full_path
+        except Exception as e:
+            self.logger.error("Failed to save %s to %s: %s", model_name, full_path, e)
+            raise
+
+    def _save_result(
+        self, result: Result, result_name: str | None, path: str | Path | None = None
+    ) -> Path:
+        """Saves a model to disk.
```
radka-j, line 46: just trying to make it clearer how this method is different from `_save_model`

```suggestion
        """Saves a result and model to disk.
```

radka-j (commented):
- [x] This is looking great! The only thought I  have is about whether there is a way to restructure the quickstart notebook so that it first just show the `ae = AutoEmulate(x, y)`  and `ae.summarise()` lines to demonstrate just how easy it is and then it gets into the various configuration options. I'm not sure what the best way to do that is. But the central message about ease of use is getting a little bit lost here.

edwardchalstrey1:
> * [x]  This is looking great! The only thought I  have is about whether there is a way to restructure the quickstart notebook so that it first just show the `ae = AutoEmulate(x, y)`  and `ae.summarise()` lines to demonstrate just how easy it is and then it gets into the various configuration options. I'm not sure what the best way to do that is. But the central message about ease of use is getting a little bit lost here.

@radka-j I've put these parts in collapsable details tags which I think works well

autoemulate/experimental/save.py
```diff
@@ -0,0 +1,156 @@
...
+                full_path = full_path / model_name
+
+        full_path.parent.mkdir(parents=True, exist_ok=True)
+        return full_path
+
+    @staticmethod
+    def _get_model_name(model):
+        """Gets the name of the model to save."""
+        model_name = model.model_name()
+        return f"{model_name}.joblib"
```
sgreenbury, line 156: Related to comment above https://github.com/alan-turing-institute/autoemulate/pull/629/files#r2225105700 this seems to add the extension?

sgreenbury:
Thanks @edwardchalstrey1 - I think this looks great. Joblib serializing looks like it works well for e.g. GPs and MLPs with the model and trained weights being reloaded. I was wondering if serializing using the torch API might be preferable to joblib (e.g. `torch.save` etc)  but I think could be an enhancement instead potentially for the future.

edwardchalstrey1:
Now should save model with `.joblib` ext correctly regardless of whether `_save_model` or `_save_result` is called

autoemulate/experimental/save.py
```diff
@@ -0,0 +1,169 @@
...
+            If None, the model's name will be used.
+        path : str, Path or None, optional
+            Path to save the model.
+            If None, the model will be saved with the model name.
+        """
+        model_filename = self._get_model_filename(model, model_name)
+        full_path = self._prepare_path(path, model_filename)
+        try:
+            if full_path.suffix != ".joblib":
+                joblib.dump(model, full_path.with_suffix(".joblib"))
```
sgreenbury, line 38: It looks like this might replace any current suffix with joblib - I think this is causing the failing test as `.pt` is a suffix in the test?

sgreenbury (approved):
Looks great - I added one comment that I think relates to the failing test and then I think this looks good to merge.

PR #632, opened 2025-07-17
Refactor Naghavi case study

### LINKED ISSUE #532: Refactor Naghavi workflow example
This needs to use `Simulator` and `Emulator` from `experimental/`.

This should happen after #529 and include #438.

### LINKED ISSUE #553: Update Naghavi workflow case study
As discussed today, the full workflow should showcase how the cardiac data is usually analysed in the literature. For example, PCA should be applied to the simulator raw outputs and that should be concatenated with other summary statistics of that output (e.g., max, min). The GP is then trained on these concatenated output features. This relates to #533.

This is low priority at the moment, opening here for visibility.

### LINKED ISSUE #637: Potential new feature: Heatmap plot for sensitivity analysis
### Summary

Display sensitivity analysis (SA) results as a heatmap, where columns correspond to model parameters and rows to features of interest for the SA.

Results can be the raw values from Sobol indexes or normalised (min - max) normalisation across features. 

### Basic Example

I'm working on the cardiovascular case study and I'm developing a figure that would looks like this:

Image

### Drawbacks

This should be a straightforward addition. I'm new to the AE codebase, so I will need some advice on maintaining the code style. I'd expect this from @radka-j .

### Unresolved questions

_No response_

### Implementation PR

#632 

### Reference Issues

_No response_

### LINKED ISSUE #648: Potential new feature: Add wave workflow to History Matching
### Summary

The history matching method `runs` only runs one wave; it would be useful to extend the functionality to:

- Have multiple waves as part of the workflow
- Allow an early stopping option with wave < n_waves if the number of emulated samples falls within the region of interest (e.g. 90%). 

### Basic Example

For an HM object `hmw`, we could:

history_matching_results = hmw.run_waves(n_waves=10, n_simulations=100, n_test_samples=10000)

history_matching_results will be a list of parameters and implausibility scores for each wave.

### Drawbacks

Needs to be well tested. 

### Unresolved questions

_No response_

### Implementation PR

This is being work at #632 

### Reference Issues

_No response_

### LINKED ISSUE #716: Add cloud sampling to HMW
Currently, in the HistoryMatchingWorkflow, we update sampling bounds based on min/max of NROY samples and then reuse latin hypercube sampling in each wave. This is inefficient as it's likely there are large regions of the space that are implausible but we still sample from them. Switch instead to cloud sampling. For each available NROY sample, sample addition points from a Gaussian with mean set to that sample value and standard deviation set to some % of that parameters range (i.e., the NROY min/max).

### LINKED ISSUE #796: Update history matching tutorials
Add plot of the simulated data (as in the calibration tutorial) and add additional plots of results.

radka-j:
Closes #532 
Closes #553 
Closes #637 
Closes #648
Closes #716 
Closes #796

Contributes to #748

### DIFF
--- a/case_studies/patient_calibration/patient_calibration_case_study.ipynb
+++ b/case_studies/patient_calibration/patient_calibration_case_study.ipynb
@@ -674,7 +674,7 @@
     "To this end, AutoEmulate implements the `HistoryMatchingWorkflow`, which can run multiple \"waves\" of history matching. In each wave:\n",
     "- new parameters are sampled from the not ruled out yet (NROY) space and evaluated on their plausability using an emulator\n",
     "- new simulations are run for a subset of those parameters\n",
-    "- the emulator is retrained with the newly simulated data\n",
+    "- the emulator is retrained with the newly simulated data from the current wave and all previous waves\n",
     "\n",
     "The `HistoryMatchingWorkflow` object takes in the simulator, the trained emulator and observations. Since the emulator was trained on a subset of the simulation parameters, we need to specify which parameters those are. `HistoryMatchingWorkflow` can also optionally be passed the data the emulator was trained on to reuse when retraining the emulator."
    ]


radka-j:
I just wanted to add a note here that this notebook is linked to on our website in the v0.3 blog post. If we choose to rename it for whatever reason, the url in the blog post needs to be updated. I wast just fixing broken links after the v1 upgrade and thought to put this note here.

radka-j:
- [ ] use the cloud technique to generate more NROY points

radka-j:
Some notes:
- there was a small bug in sensitivity analysis that has been fixed (we were plotting and using first order indices instead of total, although they are not drastically different from each other)
- I noticed that the sensitivity analysis is done on data with 8 output variables (this is data we load from file) whereas history matching etc is done on newly simulated data that only has 4 output variables (this can be seen in the cell that prints the "observations" dictionary)

radka-j:
Additionally, we should investigate the coverage of the NROY space of the generated samples
- [ ] look at GPerks `part_and_select` function (can it be made more efficient)
- [ ] look over how subsample NROY points to simulate
- [ ] look at whether we should concatenate past wave NROY points with newly generated ones

radka-j:
Note that we removed handling of constant parameters in Bayesian calibration for now. We should revisit if we want to keep this logic optionally for other use cases.

PR #638, opened 2025-07-21
Revise GP predict approach

### LINKED ISSUE #634: Experimental GP is slow to make predictions
The v0 versions handled making predictions for LOTS of samples (e.g., 35k) fine but v1 takes too long and crashes the Jupyter kernel.

sgreenbury:
Closes #634:

This PR:
- Revises the `._predict` and `.predict` methods for GPs to return batches of independent `MultivariateNormal` (`GaussianLike`) (since covariance between test data points is not required, only task covariance needs to be captured)
- Uses batches to enable prediction over large number of points at inference time
- Adds `gpytorch.settings.fast_pred_var()` for predictions

### DIFF
--- a/autoemulate/experimental/emulators/base.py
+++ b/autoemulate/experimental/emulators/base.py
@@ -281,10 +281,10 @@ class GaussianProcessEmulator(GaussianEmulator):
     """
 
     @abstractmethod
-    def _predict(self, x: TensorLike, with_grad: bool) -> GaussianProcessLike: ...
-    def predict(self, x: TensorLike, with_grad: bool = False) -> GaussianProcessLike:
+    def _predict(self, x: TensorLike, with_grad: bool) -> GaussianLike: ...
+    def predict(self, x: TensorLike, with_grad: bool = False) -> GaussianLike:
         pred = super().predict(x, with_grad)
-        assert isinstance(pred, GaussianProcessLike)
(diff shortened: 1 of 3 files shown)


radka-j (approved):
Looks great! 

I tried it with the sensitivity analysis in the Naghavi case study and it works well. I compared prediction performance between v0 and v1 for 35k samples and they are both 6 seconds.

It might be worth opening an issue to have an option to predict just the mean (doesn't need to be a v1 requirement). In some workflows that's all we need (e.g., sensitivity analysis or MCMC calibration) and in the case of making predictions for 35k samples, in v0 the prediction time goes from 6 to 0.2 seconds.

sgreenbury:
> I tried it with the sensitivity analysis in the Naghavi case study and it works well. I compared prediction performance between v0 and v1 for 35k samples and they are both 6 seconds.

That's great!

> It might be worth opening an issue to have an option to predict just the mean (doesn't need to be a v1 requirement). In some workflows that's all we need (e.g., sensitivity analysis or MCMC calibration) and in the case of making predictions for 35k samples, in v0 the prediction time goes from 6 to 0.2 seconds.

Yeah this sounds good - we have `.predict_mean_and_variance()` on the `ProbabilisticEmulator`, so a `.predict_mean()` for all emulators would be useful.

PR #639, opened 2025-07-21
Fix reaction-diffusion example and update GP predictive

### LINKED ISSUE #618: Compare performance for transformed emulator and current pipeline
In particular, the reaction diffusion example appears to have differences in performance of the final fitted emulator.

sgreenbury:
Closes #618.

This PR:
- Fixes discrepancies with v0 reaction-diffusion example
- Revises default GP parameters to those for v0
- Adds `posterior_predictive` bool to GP init that enables learned observation noise to be incorporated at inference time


radka-j (approved):
This looks great!

PR #641, opened 2025-07-22
Add str and dict serialization option for emulators and transforms in AutoEmulate class

### LINKED ISSUE #610: Refactor: implement matching from emulator and transform class instances to str

sgreenbury:
Closes #610.

This PR:
- Adds serialization/deserialization for transforms enabling transforms to be passed as they currently are in v0
- Revises short name to lower case
- Incorporates handling of either serialized or deserialized form into the main `AutoEmulate` class init
- Adds registries for emulators and transforms


sgreenbury:
> Does this mean that calling `.model_name()` on a transformed Emulator will have the original class name? If so I can remove [this change](https://github.com/alan-turing-institute/autoemulate/pull/629/files#diff-9897c8d5d5f9551a1fe2c91c9737ccba03256b70791ed3be9d248d7dfc686f59)

I think having the untransformed model name is still needed - the implementation in #629 looks good for this.

PR #621, opened 2025-07-14
Add script for benchmarking simulators with different parameters

sgreenbury:
Contributes towards #454.

PR #647, opened 2025-07-23
Add fixes from further testing and benchmarking

### LINKED ISSUE #644: Potential bug: Running GP models through compare loop sometimes triggers jitter error
### Description

I noticed that when running AutoEmulate compare loop with GP models, you sometimes get an error, but not always, which is why the tests aren't necessarily picking this up

### Reproduction steps

Run:

```python
from autoemulate.experimental.compare import AutoEmulate
from autoemulate.experimental.emulators import GaussianProcessExact
ae = AutoEmulate(x, y, models=[GaussianProcessExact])
```

### Version

I've noticed this error in the Quickstart notebook on #629 after just now merging main into this branch, so may be introduced by #639 or another recent PR

### Screenshots

Sometimes works:
Image

Sometimes throws error:
Image

### OS

Mac

sgreenbury:
Closes #644.

This PR:
- Captures the same commits as #621 but without the benchmarking since this is still in progress
- Adds fixes and functionality described in https://github.com/alan-turing-institute/autoemulate/pull/621#issuecomment-3106466642
- Adds retry logic for a given model in case the tuner returns config that results in exception upon refit [31a59d7](https://github.com/alan-turing-institute/autoemulate/pull/647/commits/31a59d78e72e18fa2c6db9b4553139db4b8ce6b5)

### DIFF
--- a/autoemulate/experimental/tuner.py
+++ b/autoemulate/experimental/tuner.py
@@ -1,6 +1,7 @@
 import json
 import logging
 
+import numpy as np
 from sklearn.model_selection import KFold
 from torchmetrics import R2Score
 
@@ -97,12 +98,6 @@ class Tuner(ConversionMixin, TorchDeviceMixin):
 
         i = 0
         while i < self.n_iter:
-            logger.debug(
-                "Tuning Model: Iteration %s: %d/%d",
-                model_class.__name__,
-                i + 1,
-                self.n_iter,
-            )
             # randomly sample hyperparameters and instantiate model
             model_config = model_class.get_random_config()
             try:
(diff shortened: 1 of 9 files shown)


radka-j (commented):
I am a bit confused about why both `AutoEmulate` and `Tuner` have `max_retries`. Should we remove `max_retries` from inside the Tuner?

autoemulate/experimental/tuner.py
```diff
@@ -94,27 +97,50 @@ def run(  # noqa: PLR0913
...
+                # Retry with new random config parameters
+                retries += 1
+                logger.warning(
+                    "Failed tuning iteration %d with model config: %s: %s",
+                    i + 1,
+                    json.dumps(model_config, default=str, separators=(",", ":")),
+                    str(e),
+                )
+                # If many retries, log error and break
+                if retries > max_retries:
```
radka-j, line 140: It feels to me like you'd never get here since the `while` is over `self.n_iter` not `max_retries`

sgreenbury: E.g. that following the `max_retries` and `attempts` in compare loop it would run https://github.com/alan-turing-institute/autoemulate/blob/5922eed740560aa8393ad54729fd69541a9f02dd/autoemulate/experimental/compare.py#L324-L329 and then move on to the next model.

radka-j (approved):
Looks great, thanks!! As discussed would be good to add a couple of comments and then it's good to merge :)

PR #650, opened 2025-07-24
Add standardization for x and y to `Emulator`

### LINKED ISSUE #645: Refactor: add support for standardize transform directly within Emulator
From discussion with @radka-j and relating to #618, an option to increase the intuitiveness for working with emulators would be to include `StandardizeTransform` within the emulators (optionally) instead of requiring `TransformedEmulator`.

This could be implemented directly in the `Emulator` within fit and predict.

The transforms included could be generalized to anything that does not break the gradients being able to be taken.

We could also extend `TransformedEmulator` to be a `nn.Module` (even with a model that is not `nn.Module`) to ensure gradients can be taken when also working with dim reduction but this could be opened as a separate non-refactor issue.

sgreenbury:
Closes #645.

This PR:
- Adds `x_transform` and `y_transform` as optional tranforms within the `Emulator` base class
- Adds API to the `__init__` of emulators to support setting this as `standardize_x` and `standardize_y` (`StandardizeTransform()`)
- Has the following defaults:
  - GPs: `standardize_x=False` (periodic functions are affected when `True`), `standardize_y=True`
  - MLPs: `standardize_x=True`, `standardize_y=True`
  - Other emulators: `standardize_x=False`, `standardize_y=False`


radka-j:
I like the default choices.

autoemulate/experimental/emulators/polynomials.py
```diff
@@ -20,6 +21,8 @@ def __init__(  # noqa: PLR0913
         self,
         x: TensorLike,
         y: TensorLike,
+        standardize_x: bool = False,
+        standardize_y: bool = False,
```
radka-j, line 25: Is it worth setting the default here to True (at least for the y) like with the MLP?

sgreenbury: From discussing, we think that having the defaults as `False` remains worthwhile so that users can opt-in to standardization and search over it in the model comparison loop still (e.g. `x_transforms_list=[[], [StandardizeTransform()]]`)

PR #658, opened 2025-07-28
Add API to skip tuning in `AutoEmulate`

### LINKED ISSUE #654: Add skip tuning mode to `AutoEmulate` class
In some cases, for the `AutoEmulate` class:
- tuning the models may not be required
- it might be useful to provide a specific model params (or default) instead

This issue is for adding this as an option to the `AutoEmulate` class.

sgreenbury:
Closes #654.

This PR:
- Adds API to `AutoEmulate` class to enable tuning to be skipped during the compare loop with either the default model params used (or given `model_kwargs`) used instead
- Updates logging to fix index and use `model_name()` method


radka-j (approved):
LGTM

PR #660, opened 2025-07-28
Add API to update simulator parameters and enable sampling when equal bounds

### LINKED ISSUE #659: Enable `Simulator` paramater_ranges to be updated and handle identical bounds
From discussion with @crangelsmith as part of #632:
- the current implementation of the `Simulator`'s parameter ranges when updated directly do not update the other related attributes (e.g. `bounds_list`) since those are set in the `__init__()`. We could introduce a getter/setter/@property for this.
- for the experimental design, when the same upper and lower bounds are present then the sampler should not be called since the value is fixed. We might need to track indices of parameters in the bounds list that will be changed and interleave with the indices that change if there is no direct API to support this.

sgreenbury:
Closes #659.

This PR:
- Adds custom setter methods for:
  - `parameter_ranges`: this updates the corresponding attributes that are set during initialization
  - `output_names`: this raises an error if new `output_names` has fewer outputs than at initialization (since this shouldn't change)
- Adds support for `LatinHypercube` sampling when parameters have equal upper and lower bounds
- Copies the `test_experimental_design.py` tests to `tests/experimental/` and updates to test equal bounds case and use torch

### DIFF
--- a/autoemulate/experimental/simulations/base.py
+++ b/autoemulate/experimental/simulations/base.py
@@ -59,6 +59,22 @@ class Simulator(ABC, ValidationMixin):
         """Dictionary mapping input parameter names to their (min, max) ranges."""
         return self._parameters_range
 
+    @parameters_range.setter
+    def parameters_range(
+        self, parameters_range: dict[str, tuple[float, float]]
+    ) -> None:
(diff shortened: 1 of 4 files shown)


crangelsmith (approved):
LGTM!

PR #686, opened 2025-07-31
Speed up tests

### LINKED ISSUE #679: Add CI testing for Python 3.11 and 3.12

EdwinB12:


autoemulate/core/tuner.py
```diff
@@ -101,7 +101,7 @@ def run(  # noqa: PLR0913
         # Initialize retries and maximum number of retries for consecutive failed tuning
         # iterations that would indicate that no model params is working for the
         # given dataset/model and that the tuning process should be stopped
-        retries, max_retries = 0, 100
+        retries, max_retries = 0, 5
```
sgreenbury, line 104: Perhaps it's worth adding an extra comment explaining the implications of the value?
```suggestion
        # max_retries is set to 5 so that if the model or available tune_params do not
        # enable model fitting, the tuner can exit. If this is an unlikely false
        # positive, the whole tuning process can be retried
        retries, max_retries = 0, 5
```

sgreenbury (approved):
Looks great, thanks @EdwinB12!

- I left a suggested comment above for the max_reties
- It looks like the permissions issue might need the permissions block within the coverage job?
- I think the docs building on push to main is ok given the reduction in time following #687 - though looking at the CI I'm not sure the notebook caching is working so perhaps that's worth opening as a new issue?

Otherwise looks good to merge!

EdwinB12:
Will merge this after removing the changes to the CI (this is done in a different PR).

PR #700, opened 2025-08-01
remove poetry support

EdwinB12:
This PR addresses: 

Remove Poetry support and any mention of poetry in docs - Resolves #366 
Add support for UV - Resolves #526 
Increase CI testing to all versions of supported python - Resolves #686 
Update workflows to remove redundant tasks

PR #702, opened 2025-08-01
681 add testing support for different os

### LINKED ISSUE #574: Add CI Support for multiple Operating Systems
Currently, our CI and release workflow only runs on Linux. We should decide what we're supporting and test on MAc and Windows too.

### LINKED ISSUE #681: Add testing support for different OS
Currently only test on linux. Should probabaly add mac to this list too.

EdwinB12:
Builds on #700 

resolves #681


EdwinB12:
Ready for Review. @radka-j @sgreenbury 

Runs tests on all OS's on all versions of python. Some specific things that needed addressing: 

1. We had a test that was designed to test catching an error when using an invalid file path (test_save.py_. What made it invalid was that is wrote to root which linux doesn't let you do). However, windows does let you do this and therefore the test didn't work across all OS's. I have replaced it with a filepath using special characters instead. 

2. Macos github runners has the unfortunate behaviour of not having the GPU available but still returned `torch.backend.mps.is_available()` still returns true. I have configured the tests to look for an environment variable. If found, then it returns DeviceMixIn returns cpu instead of mps.

PR #710, opened 2025-08-04
return dict of results from simulator

### LINKED ISSUE #709: base simulator access output as dictionary
Add a method to the base simulator class which allows to access individual outputs through their names if a list of output names is provided 

```python
    def get_outputs_as_dict(self):
        """
        Return all collected outputs as a dictionary with output names as keys
        and concatenated tensors as values.
        
        Returns:
            dict: Dictionary where keys are output names and values are tensors
                of shape [batch_size, time_steps]
        """
        if not hasattr(self, 'outputs') or not self.outputs:
            return {}
        
        output_dict = {}
        for output_name, output_list in self.outputs.items():
            if output_list:  # Check if list is not empty
                # Concatenate all batches for this output
                output_dict[output_name] = torch.cat(output_list, dim=0)
            else:
                output_dict[output_name] = None
        
        return output_dict
```

access as : 

```python 
outputs_dict = sim.get_outputs_as_dict()
interesting_one = outputs_dict["The best output"]
```

marjanfamili:
added a method to base simulator to allow access to the outputs of the simulator in a dictionary format and a cell in the simulator tutorial to explain

For instance in the battery case, the simulator can have more than 10 outputs and one would need to inspect the simulation outputs before proceeding to the emulation stage sometimes. So this would be helpful.

#709


radka-j (approved):
Looks good!

PR #713, opened 2025-08-05
Pre-commit hook to strip nb output

### LINKED ISSUE #712: Add pre-commit hook to strip notebook output

radka-j:
Closes #712


EdwinB12 (approved):
This is a great idea I think. We have docs for seeing this kind of stuff without running it.

PR #726, opened 2025-08-12
Revise transforms

### LINKED ISSUE #640: Add predict with mean only option
As mentioned in https://github.com/alan-turing-institute/autoemulate/pull/638#pullrequestreview-3042156719, in some cases only a prediction of the mean is required even when working with `ProbabilisticEmulator`.

Adding a new method to `Emulator` base class that returned the mean (`TensorLike`) could be implemented. Skipping covariance calculations would probably need to be added into the API of `predict()`.

### LINKED ISSUE #723: Revise transformed emulator prediction defaults
Following the current implementation of the transforms and the `TransformedEmulator` class there are several enhancements and revisions that would beneficial:
- Currently the `TransformedEmulator` expects `AutoEmulateTransform`s in it's list of transforms - this is unnecessary since `AutoEmulateTransform` subclasses torch `Transform`. Instead the API can be revised to the super class so that transforms already implemented in [torch](https://docs.pytorch.org/docs/stable/distributions.html#module-torch.distributions.transforms) (or [pyro](https://docs.pyro.ai/en/dev/_modules/pyro/distributions/transforms.html)) can be readily used (e.g. [`ReshapeTransform`](https://docs.pytorch.org/docs/stable/distributions.html#torch.distributions.transforms.ReshapeTransform))
- The current default path for prediction with `GaussianLike` through un-transforming involves using the analytical transformation / delta method of the Gaussian's covariance matrix when `output_from_samples=False`. This might not be optimal:
  - If a `make_positive_definite(..., clamp_eigvals=True)` is called - this might break gradients propagating and should be made more clear this would be the case with warnings
  - Some transforms do not have a clear `basis_matrix` representation so are not readily implemented for this path
  - May become computationally unefeasible and be unnecessary for some use cases (e.g. where just mean or mean + variance is required)
- The alternative prediction path for generating predictions includes a sampling approach in the transformed space and un-transforming each sample and computing the required statistic (mean / variance / covariance):
  - Currently this does not propagate gradients but we can update this path to make use of [`.rsample()`](https://docs.pytorch.org/docs/stable/distributions.html#pathwise-derivative) to propagate the gradients for arbitrary transformed distributions
  - We can add API to calculate the mean, variance, covariance as needed required so that we have API for these (`.mean` and `.variance` are not implemented with torch `TransformedDistribution` by default). We could also assign these empirical values directly to `GaussianLike` (as mean and covariance) as currently.
- Error handling should be added to handle the case when the `output_from_samples=False` but not all the `Transform`s are `AutoEmulateTransform`s (and so do not have the required methods to transform the Gaussians analytically.
- This is also discussed in #640 and the updated implementation should fit with a mean-only prediction mode

From a design perspective, the changes can also be framed as aiming to ensure the default API (with the ability to opt-in to alternative analytic / full covariance API) is:
- differentiable
- directly compatible with torch transforms
- fast/memory efficient

sgreenbury:
Closes #723.
Closes #640


autoemulate/core/compare.py
```diff
@@ -431,6 +432,112 @@ def compare(self):
...
+        model_name: str | None = None,
+        random_seed: int | None = None,
+    ) -> TransformedEmulator:
+        """
+        Create a fresh model with the best configuration and refit on new data.
+
+        This method creates a new model instance with the same configuration as the
+        best (or specified) model from the comparison, but with freshly initialized
+        parameters fitted on the provided data. This is useful for getting
+        reproducible results or avoiding randomness from the comparison workflow.
```
radka-j, line 475: This is very minor but can we be clearer here on why it's useful? Why would a user not get reproducible results if they did not refit, that seems weird.

```suggestion
        parameters fitted on the provided data. This is useful for getting
        reproducible results or avoiding randomness from the comparison workflow.
```

autoemulate/core/compare.py
```diff
@@ -431,6 +432,112 @@ def compare(self):
...
+
+        Returns
+        -------
+        TransformedEmulator
+            A new model instance with the same configuration but fresh parameters
+            fitted on the provided data.
+
+        Notes
+        -----
+        Unlike TransformedEmulator.refit() which retrains an existing model,
```
radka-j, line 497: Is this correct?
```suggestion
        Unlike TransformedEmulator.fit() which retrains an existing model,
```

autoemulate/core/compare.py
```diff
@@ -431,6 +432,112 @@ def compare(self):
...
+        this method creates a completely new model instance with reinitialized
+        parameters. This ensures that when fitting on new data that the same
+        initialization conditions are applied. This can be have an affect for example
```
radka-j, line 500:
```suggestion
        initialization conditions are applied. This can have an affect for example
```

autoemulate/core/results.py
```diff
@@ -196,3 +197,23 @@ def get_result(self, result_id: int) -> Result:
             return self._id_to_result[result_id]
         except KeyError as err:
             raise ValueError(f"No result found with ID: {result_id!s}") from err
+
+    def get_result_by_model_name(self, model_name: str) -> Result:
```
radka-j, line 201: Presumably model name here is, e.g. 'gp'. What happens if there are multiple GPs trained (e.g., with different transforms)?

autoemulate/emulators/base.py
```diff
@@ -128,6 +128,39 @@ def predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
...
+        Predict the mean of tha target variable for input `x`.
+
+        Parameters
+        ----------
+        x: TensorLike
+            Input tensor of shape `(n_samples, n_features)` for which to predict
+            the mean and variance.
+        with_grad: bool
+            Whether to compute gradients with respect to the input. Defaults to False.
+
```
radka-j, line 144: Add `n_samples` to docstring

autoemulate/emulators/base.py
```diff
@@ -128,6 +128,39 @@ def predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
...
+            Mean tensor of shape `(n_samples, n_targets)`.
+        """
+        y_pred = self._predict(x, with_grad)
+        if isinstance(y_pred, TensorLike):
+            return y_pred
+        try:
+            return y_pred.mean
+        except Exception:
+            # Use sampling to get a mean if mean property not available
+            samples = (
```
radka-j, line 157: This is very nice!

autoemulate/emulators/transformed/base.py
```diff
@@ -392,21 +401,69 @@ def _predict(self, x: TensorLike, with_grad: bool) -> OutputLike:
         if not self.output_from_samples:
             if isinstance(y_t_pred, GaussianLike):
                 return self._inv_transform_y_gaussian(y_t_pred)
-            if isinstance(y_t_pred, DistributionLike):
-                return self._inv_transform_y_distribution(y_t_pred)
-            msg = "y_pred is not TensorLike, GaussianLike or DistributionLike"
+            msg = (
+                f"y_t_pred ({type(y_t_pred)}) is not `GaussianLike` and not currently "
+                "supported for when inverting without sampling."
```
radka-j, line 406:
```suggestion
                f"Inverse transform without sampling for y_t_pred ({type(y_t_pred)}) is not currently supported, expected GaussianLike."
```

autoemulate/emulators/transformed/base.py
```diff
@@ -392,21 +401,69 @@ def _predict(self, x: TensorLike, with_grad: bool) -> OutputLike:
...
+        -------
+        tuple[TensorLike, TensorLike]
+            A tuple containing:
+            - Mean tensor of shape `(n_samples, n_targets)`.
+            - Variance tensor of shape `(n_samples, n_targets)`.
+        """
+        if not self.model.supports_uq:
+            msg = f"TransformedEmulator model ({self.model}) does not support UQ."
+            raise RuntimeError(msg)
+        y_pred = self._predict(x, with_grad)
```
radka-j, line 458: Is it ever possible that `y_pred` here is a distribution object (e.g., Gaussian) that has a mean and variance and so we do not need to use sampling? I am thinking of cases where we use `TransformedEmulator` without any transforms (or just standardisation) as is the default.

autoemulate/transforms/pca.py
```diff
@@ -60,6 +60,28 @@ def log_abs_det_jacobian(self, x: TensorLike, y: TensorLike):
         )
         raise RuntimeError(msg)
 
+    def forward_shape(self, shape):
+        """
+        Compute the forward shape transformation.
+
+        For PCA: (batch_shape..., d) -> (batch_shape..., n_components)
```
radka-j, line 67: Personally I find it clearer this way that we are maintaining all dimensions except the last one but might just be me

```suggestion
        For PCA: (batch_shape, ..., d) -> (batch_shape, ..., n_components)
```

autoemulate/transforms/pca.py
```diff
@@ -60,6 +60,28 @@ def log_abs_det_jacobian(self, x: TensorLike, y: TensorLike):
...
+        """
+        if len(shape) == 0:
+            return torch.Size([self.n_components])
+        return shape[:-1] + torch.Size([self.n_components])
+
+    def inverse_shape(self, shape):
+        """
+        Compute the inverse shape transformation.
+
+        For PCA: (batch_shape..., n_components) -> (batch_shape..., d)
```
radka-j, line 77:
```suggestion
        For PCA: (batch_shape, ..., n_components) -> (batch_shape, ..., d)
```

docs/tutorials/emulation/02_dim_reduction.ipynb
```diff
@@ -149,24 +149,39 @@
...
     "    y[train_idx],\n",
-    "    models=[\"GaussianProcess\"],\n",
+    "    models=[\"GaussianProcess\", \"GaussianProcessCorrelated\"],\n",
     "    x_transforms_list=[[StandardizeTransform()]],\n",
-    "    y_transforms_list=[[StandardizeTransform(), PCATransform(n_components=16), StandardizeTransform()]],\n",
-    "    n_splits=2,\n",
-    "    log_level=\"error\", \n",
-    ")"
+    "    y_transforms_list=[[\n",
+    "        ReshapeTransform((n, n), (n * n,)), # include torch transform to flatten first\n",
```
radka-j, line 169: I think it's really nice that we're using the ReshapeTransform from torch here, probably worth adding a note in the text above this cell to point out that torch transforms can be used.

docs/tutorials/emulation/02_dim_reduction.ipynb
```diff
@@ -149,24 +149,39 @@
    "source": [
     "from autoemulate.core.compare import AutoEmulate\n",
     "from autoemulate.transforms import *\n",
+    "from torch.distributions.transforms import ReshapeTransform\n",
     "import torch \n",
     "\n",
     "# Convert to tensors\n",
-    "x, y = torch.Tensor(X).float(), torch.Tensor(Y).float()\n",
+    "x, y = torch.Tensor(X).float(), torch.Tensor(Y).float().reshape(-1, n, n)\n",
```
radka-j, line 156: Why is the y data being reshaped here as well as in the AutoEmulate class with the ReshapeTransform?

radka-j (commented):
This is looking really great @sgreenbury!!! Just some minor comments and questions.

autoemulate/emulators/base.py
```diff
@@ -128,6 +128,39 @@ def predict(self, x: TensorLike, with_grad: bool = False) -> OutputLike:
         self._check_output(output)
         return output
 
+    def predict_mean(
+        self, x: TensorLike, with_grad: bool = False, n_samples: int = 100
+    ) -> TensorLike:
+        """
+        Predict the mean of tha target variable for input `x`.
```
radka-j, line 135:
```suggestion
        Predict the mean of the target variable for input `x`.
```

autoemulate/emulators/transformed/base.py
```diff
@@ -272,6 +276,12 @@ def _inv_transform_y_gaussian(self, y_t: GaussianLike) -> GaussianLike:
         """
         # Invert the order since the combined transform is an inversion
         for transform in self.y_transforms[::-1]:
+            if not isinstance(transform, AutoEmulateTransform):
```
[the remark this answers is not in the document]
sgreenbury: Yeah good point - as discussed it would be great to support this so the analytical path can be used by default to get mean / variance without unnecessary sampling. We'll open a new issue to explore this.

autoemulate/emulators/base.py
sgreenbury: Thanks!

autoemulate/core/results.py
sgreenbury: As discussed - it would be good to address in a new issue in more detail but for now it should:
- raise exception if more than one model with the same name for multiple transforms
- use `get_model_by_id` in the `fit_from_reinitialized` method

autoemulate/emulators/base.py
sgreenbury: Added

docs/tutorials/emulation/02_dim_reduction.ipynb
sgreenbury: The data is stored on disk / output from simulation currently flattened so this step is reshaping the ground truth back to the original shape.

autoemulate/core/compare.py
sgreenbury: As discussed - updated the name of the `AutoEmulate` `refit()` to `fit_from_reinitialized()` to make this clearer.

autoemulate/emulators/transformed/base.py
sgreenbury: This would be great, thanks for the ideas on this. Following our discussion it seems like two options going forward:
- Add analytical support fro `NormalLike` where a `GaussianLike` is transformed to `NormalLike` where each element is an independent gaussian. This would be sufficient maybe for most use cases with mean and variance.
- Look into an analytical transform of the mean when all transforms are bijective
- Look at adding an `analytical: bool` attribute on the `AutoEmulateTransform`

I'll open a new issue to capture additional development on this. It would definitely be great to have a low computational cost path for the mean and variance with gradients when it's possible analytically.

Issue #509, opened 2025-06-05
Refactor calibration dashboards


radka-j:
Refactor `HistoryMatchingDashboard` and `MCMCVisualizationDashboard` in #505 to avoid duplication of code.

PR #734, opened 2025-08-14
Add option to save plots to file

### LINKED ISSUE #732: Potential new feature: plot_sobol to file instead of terminal
### Summary

Hi, it seems the plotting functions such as `plot_sobol` plot in terminal/notebook. Would it be possible to add functionality for plotting to a file?

### Basic Example

```python
sa.plot_sobol(sobol_df, index="ST", fname="sobol_total.png")
```

or when `fname=None`, plot to terminal.

### Drawbacks

None

### Unresolved questions

_No response_

### Implementation PR

_No response_

### Reference Issues

_No response_

radka-j:
Closes #732 

Plot saving option added to:
- model performance plot generated in AutoEmulate
- Sensitivity analysis plots (Sobol and Morris)

Everywhere else we either use external plotting libraries (e.g., arviz or pure matplotlib). If we want to explore adding saving functionality to the HistoryMatchingDashboard this should be handled as part of #509.

I did consider refactoring the code to create a plot filename similar to what we do when we save models but I felt that as a user I'd prefer to choose my own plot name directly. Happy for others to disagree.

### DIFF
--- a/autoemulate/core/compare.py
+++ b/autoemulate/core/compare.py
@@ -437,7 +437,7 @@ class AutoEmulate(ConversionMixin, TorchDeviceMixin, Results):
         input_index: list[int] | int | None = None,
         output_index: list[int] | int | None = None,
         figsize=None,
-        n_cols: int = 3,
+        ncols: int = 3,
         fname: str | None = None,
     ):
         """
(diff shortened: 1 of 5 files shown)


autoemulate/core/compare.py
```diff
@@ -451,6 +452,13 @@ def plot(  # noqa: PLR0912, PLR0915
             The index of the input feature to plot against the output.
         output_index: int
             The index of the output feature to plot against the input.
+        figsize: tuple[int, int] | None
+            The size of the figure to create. If None, it is set based on the number
+            of input and output features.
+        ncols: int
```
sgreenbury, line 458:
```suggestion
        n_cols: int
```

sgreenbury: Just to note if this is updated to `ncols` https://github.com/alan-turing-institute/autoemulate/pull/734/commits/0f616ad8fa72a3a1e2275cd259aa4fbfc374842c#r2277070034 then this suggestion is obsolete.

sgreenbury (approved):
Looks great, thanks @radka-j! I made a small suggestion above that could also be implemented in a new PR if to be added.

PR #744, opened 2025-08-15
Add transformed emulator params

### LINKED ISSUE #727: Update AutoEmulate API to allow resetting TranformedEmulator args
This is based on comments from @sgreenbury in #714 

> -  Revising to not use the TransformedEmulator directly highlights we don't have API for setting this from AutoEmulate. I think it would be worth adding a new arg to AutoEmulate (perhaps transformed_emulator_kwargs: dict | None = None) - but this would be a new issue
> - Being able to set args such as posterior_predictive to non-defaults in a convenient way seems like we should revisit if we haven't an issue for it?

sgreenbury:
Closes #727.

### DIFF
--- a/autoemulate/core/compare.py
+++ b/autoemulate/core/compare.py
@@ -15,7 +15,13 @@ from autoemulate.core.plotting import calculate_subplot_layout, display_figure, 
 from autoemulate.core.results import Result, Results
 from autoemulate.core.save import ModelSerialiser
 from autoemulate.core.tuner import Tuner
-from autoemulate.core.types import DeviceLike, DistributionLike, InputLike, ModelParams
+from autoemulate.core.types import (
+    DeviceLike,
+    DistributionLike,
+    InputLike,
+    ModelParams,
+    TransformedEmulatorParams,
+)
 from autoemulate.data.utils import ConversionMixin, set_random_seed
 from autoemulate.emulators import ALL_EMULATORS, PYTORCH_EMULATORS, get_emulator_class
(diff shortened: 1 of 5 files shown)


autoemulate/core/compare.py
```diff
@@ -72,8 +79,8 @@ def __init__(  # noqa: PLR0913
             Defaults to None, in which case the data is standardized.
         model_tuning: bool
             Whether to tune the hyperparameters of the models using cross-validation.
-            If False, the models use the model_kwargs provided. Defaults to True.
-        model_kwargs: ModelParams | None
+            If False, the models use the model_params provided. Defaults to True.
+        model_params: ModelParams | None
```
radka-j, line 83: Add `transformed_emulator_params` to docstring

radka-j (approved):
Looks great!

autoemulate/core/compare.py
sgreenbury: Added.

autoemulate/core/compare.py
```diff
@@ -330,7 +339,7 @@ def compare(self):
                                     "parameters",
                                     model_cls.__name__,
                                 )
-                                best_params_for_this_model = self.model_kwargs
+                                best_params_for_this_model = self.model_params
```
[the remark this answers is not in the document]
sgreenbury: Yeah good point - it would be great to capture the default values that end up being used in the results so this is visible. I'll open an issue!

Issue #720, opened 2025-08-11
Sobol Index Calculation - Use Emulator Uncertainty if available


H-Sax:
### Description

[jrsssb_66_3_751.pdf](https://github.com/user-attachments/files/21715484/jrsssb_66_3_751.pdf)

## Current Issue with Mean-Only Approach

The current implementation uses only the GP mean prediction to calculate Sobol indices, but this misses a crucial component of uncertainty. When you only use the mean, you're assuming perfect knowledge of the true function - which defeats the purpose of having a probabilistic emulator.

## What's Missing: Emulator Uncertainty

From the paper attached, we should treat the computer model as an **unknown function** with a GP posterior distribution. This means:

- Sobol indices themselves become random variables (not fixed numbers)
- We need posterior expectations instead of point estimates:

```math
\hat{S}_i = \frac{E^*[V_i]}{E^*[\text{var}(Y)]}
```

- As a consequence there is a modification to the standard Hoeffding-Sobol decomposition:

```math
E^*[\text{var}\{z_i(X_i)\}] = \text{var}[E^*\{z_i(X_i)\}] + E[\text{var}^*\{z_i(X_i)\}]
```

### Defining the Terms:

- $z_i(x_i) = E(Y|x_i) - E(Y)$ is the **main effect** of input $X_i$
- $E^*[\cdot]$ denotes **posterior expectation** with respect to the GP
- $\text{var}^*[\cdot]$ denotes **posterior variance** with respect to the GP  
- $V_i = \text{var}\{z_i(X_i)\}$ is the **main effect variance**

### Variance Decomposition Breakdown:

1. **$\text{var}[E^*\{z_i(X_i)\}]$**: Variation of the posterior mean prediction (what current implementation captures)

2. **$E[\text{var}^*\{z_i(X_i)\}]$**: Average uncertainty within the GP posterior about what the main effect actually is (**currently ignored**)

## The Problem with Point Estimates

When using only the GP mean, you're effectively setting $E[\text{var}^*\{z_i(X_i)\}] = 0$ - assuming perfect knowledge and no emulator uncertainty. This can seriously underestimate uncertainty in sensitivity indices.

### Reproduction steps

```bash
The paper provides the formalisation on how one should properly compute Sobol indices for probablistic emulators.
```

### Version

[jrsssb_66_3_751.pdf](https://github.com/user-attachments/files/21715480/jrsssb_66_3_751.pdf)

### Screenshots

```bash
[...]

PR #747, opened 2025-08-15
Restructure tests

### LINKED ISSUE #719: Should we restructure the tests to reflect the module structure?

radka-j:
Closes #719 

This PR also updates SensitivityAnalysis to use the `predict_mean` method as it currently only uses the mean of the prediction (note that this might be amended with https://github.com/alan-turing-institute/autoemulate/issues/720). This was meant to be handled in a separate PR but got pushed here accidentally.


tests/calibration/test_history_matching.py
```diff
@@ -165,3 +159,5 @@ def test_run_max_tries():
 
     with pytest.raises(RuntimeError):
         hm.run(n_simulations=5)
+    with pytest.raises(RuntimeError):
+        hm.run(n_simulations=5)
```
radka-j, line 163:
```suggestion
```

radka-j:
There's a test failing because on the CI where loss is being recorded every step rather than every epoch. This test passes locally.

PR #751, opened 2025-08-15
Use predict_mean in sensitivity analysis

radka-j:
SensitivityAnalysis currently only uses the mean of the prediction so I've updated it to use the `Emulator.predict_mean` method. Note that this might be amended with #720


radka-j:
Closing as now part of #747

Issue #748, opened 2025-08-15
Use `fit_from_initialised` when refitting emulators in HMW


radka-j:
We have observed this leads to better emulator performance when re-fitting on new data.

radka-j:
In the Naghavi example, we have observed that refitting on all the simulated data can lead to the predictive variance getting unreasonably large. Note that in this case we are refitting on a lot of data where a lot of the rows might be quite similar and also quite a few of the columns are kept to a constant value so it might be a specific issue with this setu.

PR #752, opened 2025-08-15
Re-initialise emulator before refitting in HMW

radka-j:
Relates to #748 (to be closed once #632 is merged).

The `HistoryMatchingWorkflow` object now expects a `Result` rather than an `Emulator` object. I wasn't sure how else to make this work. It means that one needs to create an `Emulator` through `AutoEmulate` for them to be able to use this class (although they could also wrap it in a `Result` object manually if really motivated. I don't love this limitation but we have found repeatedly that re-initialising when refitting on new data is important for overall performance.


radka-j:
@sgreenbury that looks good! I'll try implementing something along those lines and then re-request a review when ready, how does that sound?

sgreenbury:
Sounds great, nice one!

PR #754, opened 2025-08-18
Add QMC Sobol sampling

### LINKED ISSUE #721: Potential new feature: Add Quasi-Monte Carlo Sampling Methods
## Problem

[1505.02350v1.pdf](https://github.com/user-attachments/files/21716191/1505.02350v1.pdf)

Currently, the package relies primarily on Latin Hypercube Sampling (LHS), but research shows **Quasi-Monte Carlo (QMC) methods consistently outperform LHS** for emulator training, MCMC, and sensitivity analysis.

### **Convergence Rates**
- **LHS**: O(N^(-0.5)) - same as Monte Carlo
- **QMC**: O(N^(-1)) - significantly faster convergence

### **Performance Comparison**
| Function Type | LHS vs MC | QMC vs MC | Winner |
|---------------|-----------|-----------|---------|
| Few dominant variables | Marginal | **Superior** | **QMC** |
| Equal importance, low interactions | Better | **Superior** | **QMC** |
| Equal importance, high interactions | Marginal | **Superior** | **QMC** |

### **Key Advantages**
- **10-100x fewer samples** for same accuracy
- **Superior space-filling**: Sobol' sequences satisfy uniformity Properties
- **Incremental sampling**: Can add new points without recomputing (unlike LHS)
- **High-dimensional performance**: Maintains efficiency as dimensions increase

### **Emulator Training**
- Better space coverage with fewer training points
- More accurate surrogate models
- Reduced computational cost

### **Bayesian History Matching**
- More efficient parameter space exploration
- Faster MCMC convergence
- Reduced clustering issues

### **Sensitivity Analysis** 
- More accurate Sobol index calculation
- Better performance for high-dimensional integration
- Consistent results across problem types

### **360-Dimensional Integration Problem**

The paper tested a mathematical function in 360 dimensions (similar to high-dimensional emulator problems):

| Method | Convergence Rate | Samples Needed for Same Accuracy |
|--------|------------------|----------------------------------|
| **Monte Carlo** | α = 0.50 | ~15,000 samples |
| **LHS** | α = 0.52 | ~15,000 samples |
| **QMC (Sobol')** | α = 0.94 | **~250 samples** |

### **Key Result**
**QMC achieved the same accuracy with 60x fewer samples** than LHS.

### **Why This Matters**
- LHS was barely better than random sampling in high dimensions
- QMC converged smoothly and monotonically 
- MC/LHS oscillated and converged slowly

This advantage becomes even more pronounced in the high-dimensional spaces common in emulator training and sensitivity analysis. The paper demonstrates that "**Quasi Monte Carlo method would appear the safest bet when integrating functions of unknown typology**" - exactly the situation we face with black-box models in emulator applications.

[1505.02350v1.pdf](https://github.com/user-attachments/files/21716257/1505.02350v1.pdf)

### Drawbacks

I think the impact is significant with minimal drawbacks. There are issues regarding the tyraining and BHM sampling regarding you break the structure of the space when one resamples are excludes sapmle points. But I think this is very much an unsolved problem. 

### Unresolved questions

_No response_

### Implementation PR

There is already a well established QMC sampler in SciPy

https://docs.scipy.org/doc/scipy/reference/stats.qmc.html

### Reference Issues

https://github.com/alan-turing-institute/autoemulate/issues/716

radka-j:
Closes #721 

This PR adds Sobol sampling from `scipy.stats.qmc` module. While implementing this, I noticed that the package also has LHS implementation and so decided to use that instead of the current implementation that relied on the `mogp-emulator` package.

### DIFF
--- a/tests/simulations/test_base_simulator.py
+++ b/tests/simulations/test_base_simulator.py
@@ -225,6 +225,7 @@ def test_update_output_names_wrong_dimension(mock_simulator):
     assert mock_simulator.output_names == ["var1", "var2"]
     assert mock_simulator.out_dim == 2
 
+
 @pytest.mark.parametrize("method", ["lhs", "sobol"])
 def test_sample(method):
     param_bouns = {
@@ -232,10 +233,10 @@ def test_sample(method):
         "param2": (10.0, 100.0),
         "param3": (1000.0, 1000.0),
     }
-    sim = MockSimulator(param_bouns, ["var1", "var2"], method)
+    sim = MockSimulator(param_bouns, ["var1", "var2"])
 
     n_samples = 1000
-    samples = sim.sample_inputs(n_samples)
+    samples = sim.sample_inputs(n_samples, method=method)
 
     assert isinstance(samples, TensorLike)
     assert samples.shape[0] == n_samples


autoemulate/simulations/base.py
```diff
@@ -151,8 +162,43 @@ def sample_inputs(
...
             set_random_seed(random_seed)  # type: ignore PGH003
-        lhd = LatinHypercube(self.param_bounds)
-        return lhd.sample(n_samples)
+
+        if len(self.sample_param_bounds) == 0:
+            # All parameters are constant - broadcast to n_samples
+            const_vals = torch.tensor(list(self.constant_params.values()))
+            return const_vals.repeat(n_samples, 1)
+
+        if method == "lhs":
```
sgreenbury, line 171: Possibly worth supporting upper and lower case?
```suggestion
        if method.lower() == "lhs":
```

autoemulate/simulations/base.py
```diff
@@ -151,8 +162,43 @@ def sample_inputs(
...
+            sampler = qmc.LatinHypercube(d=len(self.sample_param_bounds))
+        elif method == "sobol":
```
sgreenbury, line 173: As above for the case
```suggestion
        elif method.lower() == "sobol":
```

sgreenbury (approved):
This looks great, thanks @radka-j! Nice that scipy supports LHS too! I just left a few possible suggestions but looks good to merge.

PR #755, opened 2025-08-19
Add API to support plotting subset of data

### LINKED ISSUE #745: Add API to support plotting a subspace of features in output plots

sgreenbury:
Closes #745.

### DIFF
--- a/autoemulate/emulators/base.py
+++ b/autoemulate/emulators/base.py
@@ -370,6 +370,34 @@ class DeterministicEmulator(Emulator):
         assert isinstance(pred, TensorLike)
         return pred
 
+    def predict_mean_and_variance(
+        self, x, with_grad=False, n_samples=100
+    ) -> tuple[TensorLike, None]:
+        """
+        Predict the mean and variance of the target variable for input `x`.
+
+        Parameters
+        ----------
+        x: TensorLike
+            Input tensor of shape `(n_batch, n_features)` for which to predict
+            the mean and variance.
+        with_grad: bool
+            Whether to compute gradients with respect to the input. Defaults to False.
+        n_samples: int
+            Number of samples to draw when using sampling-based predictions.
+            Defaults to 100.
+
+        Returns
+        -------
+        tuple[TensorLike, None]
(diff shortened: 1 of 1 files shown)


autoemulate/core/compare.py
```diff
@@ -620,13 +629,18 @@ def plot(  # noqa: PLR0912, PLR0913, PLR0915
...
 
         # Re-run prediction with just this model to get the predictions
-        y_pred = model.predict(test_x)
-        y_variance = None
-        if isinstance(y_pred, DistributionLike):
-            y_variance = y_pred.variance
-            y_pred = y_pred.mean
+        y_pred, y_variance = (
+            model.predict_mean_and_variance(test_x)
+            if isinstance(model, ProbabilisticEmulator)
```
radka-j, line 634: Could this be simplified to a single if condition that checks if `model.supports_uq`? This should evaluate to True for both `ProbabilisticEmulator` and a `TransformedEmulator` that has a `ProbabilisticEmulator` model inside it.

radka-j (approved):
LGTM

autoemulate/core/compare.py
sgreenbury: I prefer this option though there is an issue with the typing when implemented like this (since model can be an `Emulator` and this doesn't have the `predict_mean_and_variance` method. An option could be to add the method to the `Emulator` base class and then `return mean, None` for the `DeterminsticEmulator` subclass?

sgreenbury: Implemented the change to the base class in [4402c3d](https://github.com/alan-turing-institute/autoemulate/pull/755/commits/4402c3d9bf76f52b4fbbcaf3a5ec7a31d5f3faa2) and revised the condition in c71afa3 - I think this improves things for downstream usage, what do you think?

radka-j: I was wondering about returning (mean, None) too. I think I'd be in favour of that. If the user tries to do something with a None variance it will error anyways downstream so it won't fail silently, but it allows us to write more general code.

PR #764, opened 2025-08-21
Add GP subclasses and functionality to generate new subclasses easily

### LINKED ISSUE #746: Add support for subclassing GPs easily
As discussed with @radka-j @marjanfamili @ContiPaolo  with the `neuraloperator` library for FNOs, a similar approach with partialclass to adjust the kernel and mean would be very helpful to implement so that the search could be per kernel.

sgreenbury:
Closes #746.

### DIFF
--- a/autoemulate/emulators/gaussian_process/exact.py
+++ b/autoemulate/emulators/gaussian_process/exact.py
@@ -531,10 +531,13 @@ def create_gp_subclass(
     Notes
     -----
     {name} is a subclass of {gp_base_class.__name__} and has the following parameters
-    fixed: {fixed_params_str}
+    set during initialization: {fixed_params_str}
 
-    These parameters cannot be changed during initialization and are excluded
-    from hyperparameter tuning.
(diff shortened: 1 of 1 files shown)


autoemulate/emulators/gaussian_process/exact.py
```diff
@@ -454,3 +456,115 @@ def forward(self, x):
...
+}
+
+
+def create_gp_subclass(
+    name: str, gp_base_class: type[GaussianProcess], **fixed_kwargs
+) -> type[GaussianProcess]:
+    """
+    Create a subclass of GaussianProcess with given fixed_kwargs.
+
+    This function creates a subclass of GaussianProcess where certain parameters
```
radka-j, line 474: I think we should be careful with the language here - this excludes the param from the AutoEmulate tuning (in that you set a kernel) but there is still the GPyTorch tuning of the kernel hyperparameters that happens. Reading this would make me think that the GPyTorch tuning is being turned off.

radka-j (approved):

PR #765, opened 2025-08-21
Update `BayesianCalibation`to optionally include model variance and discrepancy

### LINKED ISSUE #760: Update uncertainty quantification in Bayesian calibration
From the discussion with @radka-j:
- Currently the Bayesian calibration works through specification of:
   - A number of real-world observations
   - Some noise/uncertainty associated with those observations (e.g. measurement uncertainty)
- Since there are some contexts in which the measurement uncertainty will be very small / ~zero, it would be beneficial to be able to express this setting in the calibration
- Additionally, the uncertainty we consider at the moment is only this observation uncertainty - analogously to the history matching workflow, it might also make sense to include other uncertainty:
   - UQ learned in the model
   - Additional uncertainty from uncertainty in the model choice (optional)

This issue aims to introduce API to support including these additional two uncertainties. Following this update it would be helpful to explore further using the workflow/API in application to level sets (parameters that generate at least/most a given level of the output).

sgreenbury:
Closes #760.

This PR:
- Adds API for model variance and discrepancy
- Updates likelihood calculation to (optionally) include model variance and discrepancy

### DIFF
--- a/autoemulate/calibration/bayes.py
+++ b/autoemulate/calibration/bayes.py
@@ -199,49 +199,32 @@ class BayesianCalibration(TorchDeviceMixin):
             # - add model_discrepancy (variance)
             # - model variance (variance, if specified and provided by emulator)
 
-            # Get observation noise and ensure a 2D Tensor
-            obs_noise = self.observation_noise[output]
-            if isinstance(obs_noise, torch.Tensor) and obs_noise.ndim == 0:
-                obs_noise = obs_noise.reshape(1, 1).to(self.device)
-            elif isinstance(obs_noise, torch.Tensor) and obs_noise.ndim != 0:
-                msg = f"Observation noise tensor for output ({output}) must be scalar."
-                raise ValueError(msg)
-            elif isinstance(obs_noise, float):
-                obs_noise = torch.Tensor([[self.observation_noise[output]]])
-            else:
-                msg = "Observation noise must be a float or 0D tensor."
-                raise ValueError(msg)
-            assert isinstance(obs_noise, TensorLike)
-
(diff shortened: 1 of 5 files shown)


autoemulate/calibration/bayes.py
```diff
@@ -41,7 +43,15 @@ def __init__(  # noqa: PLR0913
         observations: dict[str, TensorLike]
             A dictionary of observations for each output.
         observation_noise: float |  dict[str, float]
-            A single value or a dictionary of values (one per output). Defaults to 0.1.
+            A single value or a dictionary of values (one per output) of measurement
+            noise measured in terms of standard deviation. Defaults to 0.1.
+        model_uncertainty: bool
+            Whether to include the variance associated with model predictions when
+            calculating the likelihood. Defaults to False.
+        model_discrepancy:
```
radka-j, line 51: if we are asking the user to specify model discrepancy as variance, then maybe we should specify observation noise as a variance too? Either way, it would be better if these are consistent with each other.

We should similarly specify what the expected unit is in the history matching dosctrings (currently it's unclear but in the variable names we treat everything is a variance).

sgreenbury: Yeah agree it would be good to have this consistency - it seems like treating all as variance might be most intuitive since we can add these up? I'll update the history matching docstring.

sgreenbury: Variance is updated in [4b313a1](https://github.com/alan-turing-institute/autoemulate/pull/765/commits/4b313a1afece2beb3eb1970549a37e71bf7e0838).

PR #766, opened 2025-08-22
should report failed doc builds

EdwinB12:


EdwinB12:
@radka-j - Could you give this a super quick scan please to make me feel better about merging it

docs/tutorials/emulation/01_quickstart.ipynb
```diff
@@ -105,7 +105,7 @@
    "outputs": [],
    "source": [
     "# Run AutoEmulate with default settings\n",
-    "ae = AutoEmulate(x, y, log_level=\"error\")"
+    "ae = AutoEmulate(x, y, log_level=\"error\", n_iter=3)"
```
radka-j, line 108: I feel it would be better if the Quickstart wasn't using any custom settings (setting the log level is fine), the point here is to show that one can use `AutoEmulate` as is without having to worry about setting anything.

EdwinB12: Okay fair enough

radka-j (approved):

PR #768, opened 2025-08-22
Set random seeds in Bayes tutorial for reproducibility

radka-j:
I added random seeds to the simulator sampling and emulator fitting to hopefully get consistent results in the Bayesian calibration tutorial.


EdwinB12 (approved):
Looks great. 

Should we fix these overlapping plots whilst we are here? 

image

PR #778, opened 2025-08-28
double pendulum simulator

marjanfamili:
The simulator is correct and produce correct results, however GP is not fitting that well for it.

A notebook in exploratory is demonstrating the simulator running correctly and training GP.

### DIFF
--- a/autoemulate/experimental/exploratory/double_pendulum_simulator.ipynb
+++ b/autoemulate/experimental/exploratory/double_pendulum_simulator.ipynb
@@ -1,0 +1,135 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
+    "\n",
+    "import arviz as az\n",
+    "\n",
+    "from autoemulate.simulations.double_pendulum import DoublePendulum\n",
+    "from autoemulate.core.compare import AutoEmulate\n",
+    "from autoemulate.calibration.bayes import BayesianCalibration\n",
+    "from autoemulate.emulators import GaussianProcess\n",
+    "from autoemulate.transforms import PCATransform\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "simulator = DoublePendulum()\n",
+    "x = simulator.sample_inputs(100, random_seed=42)\n",
+    "y = simulator.forward_batch(x)\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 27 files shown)


marjanfamili:
With mixed kernel and more sim data (500) the accuracy of training can be up to 76% and test up to 60%

marjanfamili:
This could be grounds for exploring further with history matching or active learning

autoemulate/experimental/exploratory/double_pendulum_simulator.ipynb
```diff
@@ -0,0 +1,205 @@
...
+    "from sklearn.gaussian_process.kernels import ExpSineSquared\n",
+    "import numpy as np\n",
+    "\n",
+    "# Periodic kernel for oscillatory behavior + RBF for chaos\n",
+    "kernel = RBF(length_scale=1.0) * ExpSineSquared(length_scale=1.0, periodicity=2*np.pi) + WhiteKernel()\n",
+    "\n",
+    "ae = AutoEmulate(x, y_new, \n",
+    "    models=[GaussianProcess],\n",
+    "    model_params=[{\"kernel\": kernel}],\n",
+    "    y_transforms_list=[[PCATransform(n_components=8)]], log_level=\"progress_bar\"\n",
```
radka-j, line 104: This is very unintuitive but you need to pass `model_tuning=False` otherwise the model params get ignored and tuning happens anyways. The `model_params` should be just the dictionary, not a list. You might also want to add `'standardize_y': True` as it usually improves performance (I thought this would happen by default but it doesn't seem to be). So it should look something like

```python
ae = AutoEmulate(x, y_new, 
    models=[GaussianProcess],
    model_params={
        "kernel": kernel,
        'standardize_y': True
    },
    model_tuning=False,
    y_transforms_list=[[PCATransform(n_components=8)]], 
    log_level="progress_bar"
)
```

radka-j: I have a look at how the other kernels are specified in AutoEmulate and there's a bit more going on. In this case it would look something like:

```python
def kernel(
    n_features: int | None, n_outputs: torch.Size | None
 ) -> Kernel:
    rbf_kernel = RBF(
        ard_num_dims=n_features, 
        batch_shape=n_outputs
    ).initialize(lengthscale=torch.ones(n_features))
    exp_sine_kernel = ExpSineSquared(
        periodicity=2*np.pi, 
        ard_num_dims=n_features, 
        batch_shape=n_outputs
    ).initialize(lengthscale=torch.ones(n_features)) 
    return rbf_kernel * exp_sine_kernel + WhiteKernel(ard_num_dims=n_features, batch_shape=n_outputs)
```

Max mention the `ard_num_dims` but I am not sure and it is not clear to me what the difference between calling `initialise()` vs specifying `lenghscale` in the args as you do is. As far as I can tell your kernel specification works as well this but just thought I'd mention it.

marjanfamili: Thank you this is very helpful

PR #779, opened 2025-08-28
add random seed into bayesian tutorial

EdwinB12:
This is mainly motivated to reduce the urn time of the bayesian calibration notebook


EdwinB12:
we may want to add this as an option into the API rather than setting it globally outside of the package.

EdwinB12:
I prefer it outside of the package if I'm honest. I think its much more transparent for the user. Especially as the advised way to set the seed is a general random seed across all of Pyro.

If we could set a seed for individual stochastic elements, then that would be different and I would be more in favour of including it in the AE API.

radka-j:
I'm happy for this to be merged as is without updating the API.

Separate to this PR, I'd be interested to hear more about your thinking behind this and whether it extends to how we handle the other random seeds in the package?

radka-j (approved):

Issue #466, opened 2025-05-14
Refactor: Review hyperparameter tuning


radka-j:
Check that all the defaults are sensible options. Are there cases where the options should/can be data driven in some way?

radka-j:
For example, should we search over more GP kernels (or more/different types of kernel compositions)?

radka-j:
Another option is to look at more sophisticated hyperparameter tuning, for example, using [Raytune](https://docs.ray.io/en/latest/tune/index.html) or [Optuna](https://github.com/optuna/optuna).

edwardchalstrey1:
Some additional suggestions (from co-pilot):
- Expand the kernel search space: Add more base kernels (e.g., `Periodic`, `DotProduct`, `WhiteNoise`) and more compositions (nested sums/products).
    ```
        from gpytorch.kernels import (
            DotProductKernel,
            PeriodicKernel,
            ScaleKernel,
            WhiteNoiseKernel,
        )
     ```
- Automate kernel composition: To automate kernel composition in a PyTorch/GPyTorch-based Gaussian Process, you can implement a kernel grammar search or greedy composition search.
- Use Bayesian optimization: Replace random search in [Tuner](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) with Bayesian optimization (e.g., [Optuna](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), [GPflowOpt](https://gpflowopt.readthedocs.io/en/latest/intro.html).
- Early stopping: Use validation performance to prune poor kernel choices early.

edwardchalstrey1:
So.. we might want to define a "grammar" of "base kernels" which I guess for AutoEmulate would want to be as broad as possible

radka-j:
Yes I think this is something we might want to consider

> So.. we might want to define a "grammar" of "base kernels" which I guess for AutoEmulate would want to be as broad as possible

radka-j:
I think we should also consider whether a kernel is strictly a "hyperparameter". Each kernel has its own hyperparameters that get optimised when fitted with GPytorch. I think we should instead consider each kernel a different emulator as the user might find it useful to get results for each one rather than just the "best performing" one.

radka-j:
[...]

PR #780, opened 2025-08-29
Don't tune if user passes model_params

radka-j:
Currently, if the user passes `model_params` but does not specify `model_tuning=False` then the passed params get ignored. This can lead to user error (this has happened). It feels that the expected behaviour should be that in that case no tuning happens. This PR proposes this change. An alternative would be to raise an error or at least a warning, that would also be a reasonable option.

### DIFF
--- a/autoemulate/core/compare.py
+++ b/autoemulate/core/compare.py
@@ -159,6 +159,7 @@ class AutoEmulate(ConversionMixin, TorchDeviceMixin, Results):
         self.n_bootstraps = n_bootstraps
         self.max_retries = max_retries
         self.model_params = model_params or {}
+        self.model_tuning = model_params is None
         self.transformed_emulator_params = transformed_emulator_params or {}
 
         # Set up logger and ModelSerialiser for saving models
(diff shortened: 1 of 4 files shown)


sgreenbury:
Thanks for the PR fixing this, adding a few comments/thoughts though feel free to merge too as I think this looks great:
- I think logging a warning would be good as you mentioned, I added a suggestion above
- An alternative option might be to remove `model_tuning` entirely from the API to further simplify. When `model_params = None`, run tuning otherwise use `model_params` with no tuning.
- The API currently doesn't support specifying some `model_params` and while still tuning others (including non-tunable ones like e.g. in [02_dim_reduction](https://alan-turing-institute.github.io/autoemulate/tutorials/emulation/02_dim_reduction.html) specify in (`posterior_predictive=True`). Since this is not supported by the current implementation this might make most sense to address as part of #466 or in a new issue?

radka-j:
I like the idea of removing `model_tuning` entirely. Agree that we can revisit more complex use cases (e.g., tune some models but not others) as part of #466, there's a lot to consider there.

radka-j:
I guess removing `model_tuning` doesn't allow the user to just use the default params without tuning which seems like a nice option to have.

PR #810, opened 2025-09-05
Github action to pull data and write out to the branch. Also link in …

EdwinB12:
…the readme

The action is only triggered when on main. So we can either YOLO this and merge this entire PR. Or we could: 

Push just the workflow to main. This won't trigger straight away. I will manually trigger it and then check it creates the report as designed. We then add the readme afterwards with a link afterwards in a seperate PR so we don't have a deadlink on the repository for a few minutes. . 

Also, What badge should I use for this?


radka-j (approved):
YOLO? haha

As for the badge, I think you can make custom ones with text you want? We could do a single colour one like the `documentation` badge that just sets something like "GitHub stats"

PR #820, opened 2025-09-08
Add dataset for advection-diffusion example

### LINKED ISSUE #803: Add advection-diffusion (shallow water equations) dataset/simulator
Add advection-diffusion simulator which solves the shallow water equations as additional spatio-temporal test case, similarly to #788

ContiPaolo:
Closes #803 

This PR:
- Adds `AdvectionDiffusion` simulator, mimicking the `ReactionDiffusion` structure in #794
- Adds experimental notebook for the advection diffusion dataset

### DIFF
--- a/autoemulate/simulations/advection_diffusion.py
+++ b/autoemulate/simulations/advection_diffusion.py
@@ -1,0 +1,243 @@
+import numpy as np
+import scipy.sparse as sp
+import torch
+from scipy.fft import fft2, ifft2
+from scipy.integrate import solve_ivp
+
+from autoemulate.core.types import NumpyLike, TensorLike
+from autoemulate.simulations.base import Simulator
+
+integrator_keywords = {}
+integrator_keywords["rtol"] = 1e-6
+integrator_keywords["method"] = "RK45"
+integrator_keywords["atol"] = 1e-8
+
+
+class AdvectionDiffusion(Simulator):
+    """Simulate the 2D vorticity equation (advection-diffusion)."""
+
+    def __init__(
+        self,
+        parameters_range: dict[str, tuple[float, float]] | None = None,
+        output_names: list[str] | None = None,
+        return_timeseries: bool = False,
+        log_level: str = "progress_bar",
+        n: int = 50,
+        L: float = 10.0,
+        T: float = 80.0,
+        dt: float = 0.25,
+    ):
+        """
+        Initialize the AdvectionDiffusion simulator.
+
+        Parameters
+        ----------
+        parameters_range: dict[str, tuple[float, float]]
+            Mapping of input parameter names to (min, max) ranges.
+        output_names: list[str]
+            List of output parameter names.
+        log_level: str
+            Logging level for the simulator.
+        return_timeseries: bool
+            Whether to return the full timeseries or just the final snapshot.
+        n: int
+            Number of spatial points per direction.
+        L: float
+            Domain size in X and Y directions.
+        T: float
+            Total simulation time.
+        dt: float
+            Time step size.
+        """
+        if parameters_range is None:
+            parameters_range = {
+                "nu": (0.0001, 0.01),  # viscosity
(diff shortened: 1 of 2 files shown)


autoemulate/simulations/advection_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+            n_time = int(self.T / self.dt)
+            y_reshaped = y.reshape(y.shape[0], n_time, self.n, self.n, 1)
+        else:
+            y_reshaped = y.reshape(y.shape[0], 1, self.n, self.n, 1)
+
+        return {
+            "data": y_reshaped,
+            "constant_scalars": x,
+            "constant_fields": None,
+        }
```
sgreenbury, line 95: Since this method is also in the reaction-diffusion simulator - perhaps we should move this to a new `SpatioTemporalSimulator(Simulator)` base class? For example, something like:
```python
class SpatioTemporalSimulator(Simulator):
    """Base class for spatiotemporal simulators.

    Attributes
    ----------
    return_timeseries: bool
        Whether to return the full timeseries or just the final snapshot.
    n: int
        Number of spatial points per direction.
    L: float
        Domain size in x and y directions.
    T: float
        Total simulation time.
    dt: float
        Time step size.
    """

    return_timeseries: bool
    n: int
    L: float
    T: float
    dt: float

    def forward_samples_spatiotemporal(
        self, n: int, random_seed: int | None = None
    ) -> dict:
        """Reshape to spatiotemporal format and return data plus constants."""
        x = self.sample_inputs(n, random_seed)

        y, x = self.forward_batch(x)

        if self.return_timeseries:
            n_time = int(self.T / self.dt)
            y_reshaped = y.reshape(y.shape[0], n_time, self.n, self.n, 1)
        else:
            y_reshaped = y.reshape(y.shape[0], 1, self.n, self.n, 1)

        return {
            "data": y_reshaped,
            "constant_scalars": x,
            "constant_fields": None,
        }


class AdvectionDiffusion(SpatioTemporalSimulator): ...

class ReactionDiffusion(SpatioTemporalSimulator): ...
```

It might be worth considering the structure (e.g. variable names, methods to add, etc) of the `SpatioTemporalSimulator` simulator a bit more - either adding an initial version here or leaving it entirely for a new issue/PR seem good to me.

autoemulate/simulations/advection_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+from autoemulate.core.types import NumpyLike, TensorLike
+from autoemulate.simulations.base import Simulator
+
+integrator_keywords = {}
+integrator_keywords["rtol"] = 1e-6
+integrator_keywords["method"] = "RK45"
+integrator_keywords["atol"] = 1e-8
+
+
+class AdvectionDiffusion(Simulator):
```
sgreenbury, line 16: I think it will be interesting to see the the predictive performance of AE with this system. Perhaps we should open an issue to add to the dim reduction tutorial? Or a new tutorial if different features of AE could be useful to demonstrate on this dataset (e.g. a different transform).

sgreenbury (approved):
This looks great, thanks for adding this @ContiPaolo! I left a few comments/ideas that could be revisited in new issues/PRs so this can be merged as is if preferred.

autoemulate/simulations/advection_diffusion.py
```diff
@@ -0,0 +1,243 @@
...
+    """Simulate the 2D vorticity equation (advection-diffusion)."""
+
+    def __init__(
+        self,
+        parameters_range: dict[str, tuple[float, float]] | None = None,
+        output_names: list[str] | None = None,
+        return_timeseries: bool = False,
+        log_level: str = "progress_bar",
+        n: int = 50,
+        L: float = 10.0,
```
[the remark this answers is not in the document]
ContiPaolo: Yes, it is the length of the square domain, so I would say `float` is more appropriate

PR #825, opened 2025-09-10
add dtype to dataset_class

marjanfamili:
Some models would expect specific dtype so its best to handle this in the patio temporal data class 

This PR adds an attribute dtype for the data class and forces all numerical values to take this specific dtype


sgreenbury (approved):
Looks great, thanks @marjanfamili!

PR #840, opened 2025-09-16
Add initial emulator base class for The Well

### LINKED ISSUE #828: Explore wrapping the well models, metrics, dataset and utils
Initial notes (WIP):
- [x] Metrics: these are importable (like VMRSE)
- [x] Models: these are importable provided done with:
```python
from the_well.benchmark import models
fno = models.FNO
```
- [x] Data structure for models: the trainer rearranges, assumes time in channel dim, we can add attribute on emulator that says whether this is expected. See [tutorial]().
- [x] Wrapping The Well: create a base class that has model attribute. This can be set at init. Subclasses can fix this. Similar to GP subclassing currently impled.
- [x] Training: is handled in separate module. To use with fit, we can wrap a trainer in the Well base class.
- [x] Autoregressive training: how/what is implemented here in the well?
   - No autoregressive training - next m time-step prediction
- [x] Autoregressive inference: how/what is implemented here in the well?
  - A rollout model is [implemented here](https://github.com/PolymathicAI/the_well/blob/a123419f8886f70660ace1438f792787a559d246/the_well/benchmark/trainer/training.py#L212-L257) with [AR predictions](https://github.com/PolymathicAI/the_well/blob/a123419f8886f70660ace1438f792787a559d246/the_well/benchmark/trainer/training.py#L249-L253) in the form of a moving batch.

sgreenbury:
Closes #828.


radka-j:
Closing as we won't be adding this to AE

PR #846, opened 2025-09-19
Update default models

### LINKED ISSUE #534: Refactor GP to return result per kernel
Currently we treat the kernel as a hyperparameter but really we want to treat each as  a separate model and report results for each kernel.

### LINKED ISSUE #843: Update default emulator model set
I'd propose we limit the default emulator set to just PyTorch models. If the users are interested in also training the sklearn models, they have to choose to do this explicitly.

### LINKED ISSUE #844: Add subclasses for correlated GP
We've done this for GPs, why not the correlated GP?

Also relates to discussion in #534

### LINKED ISSUE #858: Revisit GP kernel composition
At the moment we have a few example kernels that are compositions of other kernels. But really, any two kernels we have are composable with each other. I'd be in favour of adding linear and constant kernel methods and then adding a tutorial on how to compose kernels to create new GP classes (relates to #435). 

As part of this I'd suggest that we only keep the non-composite kernels as the emulators to use by default   in AutoEmulate (related to #843).

radka-j:
Closes #534
Closes #843 
Closes #844 
Closes #858

## Overview of new default behaviour:
- only loop over PyTorch models (unless user specifies otherwise by setting `pytorch_only=False`)
- for GPs, only use the non-composite kernels (unless user specifies composites they want to test as well)

This PR also adds a new tutorial that demonstrates how to add new composite GP kernels as well as new emulators through subclassing.

## Questions:
- should the GP subclasses also specify a mean module or we leave the choice to tuning?
- for tuning, is it worth always testing the default options as well as a few random choices? the defaults are often sensible and it would be good to  make sure we don't skip. I am thinking here especially of GPs (and testing the constant mean option)
- we now have a large number of models we tune - can we reduce this (e.g., drop correlated GPs from the default set)?

## TODO:
- [ ] introduce default set of emulators (not all, exclude correlated GPs) and explain in notebook
- [ ] check grad behaviour with subclasses (is there an issue with having removed the constant mean default?)


radka-j:
Closing as decided to split into 2 PRs.

PR #863, opened 2025-09-25
Explore extending `TheWellEmulator` API

sgreenbury:
Closes #855.
Closes #856.

(builds on changes in #840)

This is WIP exploration of adding further API to `TheWellEmulator` base class that would support more complex training and losses.


radka-j:
Closing as we won't be adding this to AE

PR #865, opened 2025-09-25
Update default models

### LINKED ISSUE #534: Refactor GP to return result per kernel

### LINKED ISSUE #843: Update default emulator model set

### LINKED ISSUE #844: Add subclasses for correlated GP

### LINKED ISSUE #858: Revisit GP kernel composition

radka-j:
Closes #534 
Closes #843 
Closes #844 
Closes #858


autoemulate/core/compare.py
```diff
@@ -186,12 +191,18 @@ def probablistic_emulators() -> list[type[Emulator]]:
         return [emulator for emulator in ALL_EMULATORS if emulator.supports_uq]
 
     @staticmethod
-    def list_emulators() -> pd.DataFrame:
+    def list_emulators(subset: str = "default") -> pd.DataFrame:
```
sgreenbury, line 194: Perhaps it would be worth updating the type hint to show the two supported values?
```suggestion
    def list_emulators(subset: Literal["all", "default"] = "default") -> pd.DataFrame:
```

PR #871, opened 2025-10-03
Update GP factory

### LINKED ISSUE #867: Implement an approach for getting default params of a class
It would be helpful to have a way of getting default params from a class directly. Inspecting an init is limited fro GP sublcasses constructed through the factory `create_gp_sublcass`. Relates to #249.

radka-j:
Closes #867


sgreenbury (approved):
Looks great, thanks @radka-j! I had:
- one comment above that would be a separate issue
- a suggestion opened in #873 to capture a change outside of the changed code to ensure `covar_module_fn` and `mean_module_fn` get added to the subclass's docstring. I also made a minor change to how all the fixed params formatted there (as well as minor formatting change for fixed params)

PR #883, opened 2025-10-09
adding general torchmetrics support

EdwinB12:


autoemulate/core/compare.py
```diff
@@ -120,13 +127,25 @@ def __init__(
             it will show a progress bar during model comparison. It will set the
             logging level to "error" to avoid cluttering the output
             with debug/info logs.
+        tuning_metric: str | TorchMetrics
+            Metric to use for hyperparameter tuning. Can be a string shortcut
+            ("r2", "rmse", "mse", "mae") or a MetricConfig object. Defaults to "r2".
+        evaluation_metrics: list[str | TorchMetrics] | None
```
radka-j: It looks like we use the first metric in this list for selecting the best performing model, can we add this to the docs to make it clear this is the case (and log the choice as well)?

autoemulate/core/metrics.py
```diff
@@ -0,0 +1,154 @@
...
+    ----------
+    metric : MetricLike
+        The torchmetrics metric class or partial.
+    name : str
+        Display name for the metric.
+    maximize : bool
+        Whether higher values are better. Defaults to True.
+    """
+
+    metric: MetricLike
```
sgreenbury: As for: https://github.com/alan-turing-institute/autoemulate/pull/883/files#r2444380674
```suggestion
```

autoemulate/core/metrics.py
```diff
@@ -0,0 +1,154 @@
...
+        self,
+        metric: MetricLike,
+        name: str,
+        maximize: bool,
+    ):
+        self.metric = metric
+        self.name = name
+        self.maximize = maximize
+
+
```
sgreenbury: Related to https://github.com/alan-turing-institute/autoemulate/pull/883/files#r2444396838, this could then be implemented for the subclass moving parts of model_selection module into the class, e.g.
```suggestion
    def __call__(self, y_pred: OutputLike, y_true: TensorLike) -> TensorLike:
        """Calculate metric."""
        if not isinstance(y_pred, TensorLike):
            raise ValueError(f"Metric not implemented for y_pred ({type(y_pred)})")
        if not isinstance(y_true, TensorLike):
            raise ValueError(f"Metric not implemented for y_true ({type(y_true)})")

        metric = self.metric()
        metric.to(y_pred.device)
        # Assume first dim is a batch dim, flatten others for metric calculation
        metric.update(y_pred.flatten(start_dim=1), y_true.flatten(start_dim=1))
        return metric.compute()
```

EdwinB12:
Thanks Both for comments! Aslong a tests pass, this should be good to go.

Issue #884, opened 2025-10-09
Add metrics for probabilistic outputs


sgreenbury:
Implementing the following probabilistic metrics would be useful to enable the performance of an emulator's UQ to be quantified.

As mentioned in #517, the following could be good to start:
- [ ] CRPS
- [x] Coverage metrics (e.g. proportion times interval captures GT)

sgreenbury:
> See https://lightning.ai/docs/torchmetrics/stable/regression/crps.html

This looks great for 1D target and for a tensor of predictive samples. I wonder if we can also generalize to:
- `y_true` when 2D
- `DistributionLike` as the `y_pred` type

PR #897, opened 2025-10-16
Add calibration plot

radka-j:
Contributes to #884

### DIFF
--- a/docs/tutorials/emulation/01_quickstart.ipynb
+++ b/docs/tutorials/emulation/01_quickstart.ipynb
@@ -285,7 +285,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ae.plot_calibration(best.model, figsize=(6, 4))"
+    "ae.plot_calibration(best.model)"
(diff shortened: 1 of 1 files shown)


autoemulate/core/plotting.py
```diff
@@ -412,8 +411,143 @@ def create_and_plot_slice(
...
+    if figsize is None:
+        figsize = (6, 6)
+    fig, ax = plt.subplots(figsize=figsize)
+
+    if empirical.shape[1] == 1:
+        ax.plot(levels, empirical, marker="o", label="empirical")
+    else:
+        # multiple outputs: plot each dimension
+        for i in range(empirical.shape[1]):
+            ax.plot(levels, empirical[:, i], marker="o", label=f"dim {i}")
```
sgreenbury: Perhaps we could indicate its for the output in the label, e.g.:
```suggestion
            ax.plot(levels, empirical[:, i], marker="o", label=f"$y_{i}$")
```

PR #904, opened 2025-10-29
Add CRPS metric

### LINKED ISSUE #884: Add metrics for probabilistic outputs

sgreenbury:
Contributes towards #884.

This PR:
- Adds `ProbabilisticMetric` base class (no new methods for the moment, but allows extension)
- Adds CRPS metric with handling for `OutputLike`
- Updates the `Metric` and `evaluate()` API to support passing `OutputLike`
- Revises model selection to use `.predict()` so that probabilistic or deterministic metrics can be used interchangeably
- Adds einops dep

This might close #884 if no other probabilistic metrics are planned for the now.

### DIFF
--- a/autoemulate/core/model_selection.py
+++ b/autoemulate/core/model_selection.py
@@ -156,7 +156,7 @@ def bootstrap(
     x: TensorLike,
     y: TensorLike,
     n_bootstraps: int | None = 100,
-    n_samples: int = 100,
+    n_samples: int = 1000,
     device: str | torch.device = "cpu",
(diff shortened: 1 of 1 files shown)


EdwinB12 (approved):
Looks great Sam!

PR #907, opened 2025-10-30
Update plots

radka-j:
This PR:

1. adds option to plot features vs outputs using filled error regions rather than error bars (which are still the default)
    - this was the original version of the plot which we removed as we thought it was misleading but it would be nice to have the option to make this plot in some use cases (I have seen it elsewhere)
    - enable by setting `error_style=fill`
 
image

2. adds plot of true vs predicted values for each output (not by feature as the plot above)

image

3. For both plots the user has the option to pass in names of the input/output features for labelling. This is now demonstrated in the quickstart notebook

image

### DIFF
--- a/autoemulate/core/compare.py
+++ b/autoemulate/core/compare.py
@@ -2,6 +2,7 @@ import inspect
 import warnings
 from datetime import datetime
 from pathlib import Path
+from typing import Literal
 
 import joblib
 import matplotlib.pyplot as plt
@@ -597,7 +598,7 @@ class AutoEmulate(ConversionMixin, TorchDeviceMixin, Results):
         output_index: list[int] | int | None = None,
         input_ranges: dict | None = None,
         output_ranges: dict | None = None,
-        error_style: str = "bars",
+        error_style: Literal["bars", "fill"] = "bars",
         figsize=None,
         ncols: int = 3,
         fname: str | None = None,


autoemulate/core/plotting.py
```diff
@@ -93,18 +97,46 @@ def plot_xy(
...
-            elinewidth=2,
-            capsize=3,
-            alpha=0.5,
-            # use unicode for sigma
-            label="pred. (±2\u03c3)",
-        )
+        if error_style not in ["bars", "fill"]:
+            msg = "error_style must be one of ['bars', 'fill']"
+            raise ValueError(msg)
+        if error_style == "bars":
```
sgreenbury: Relates to above suggestion on case sensitivity - could also have this assigned once earlier in the function?
```suggestion
        if error_style.lower() == "bars":
```

autoemulate/core/compare.py
```diff
@@ -765,6 +796,116 @@ def subset_outputs(x, y, y_p):
...
+        test_x, test_y = self._convert_to_numpy(test_x, test_y)
+        assert test_x is not None
+        assert test_y is not None
+        assert y_pred is not None
+        y_pred, _ = self._convert_to_numpy(y_pred, None)
+        test_x = self._ensure_numpy_2d(test_x)
+        test_y = self._ensure_numpy_2d(test_y)
+        y_pred = self._ensure_numpy_2d(y_pred)
+
+        # Figure out layoyt
```
sgreenbury:
```suggestion
        # Figure out layout
```

autoemulate/core/plotting.py
```diff
@@ -93,18 +97,46 @@ def plot_xy(
...
+                alpha=0.5,
+            )
+        else:
+            ax.fill_between(
+                x_sorted,
+                y_pred_sorted - 1.96 * y_std,
+                y_pred_sorted + 1.96 * y_std,
+                color=pred_points_color,
+                alpha=0.2,
+                label="0.95 CI",
```
sgreenbury: Perhaps for consistency with use of $2\sigma$ elsewhere in the plotting?
```suggestion
                label="±2\u03c3",
```

PR #918, opened 2025-11-04
Update ensemble init update kwargs

### LINKED ISSUE #914: Potential bug: fit_from_reinitialized when used with EnsembleMLP raises error
### Description

When fitting the `EnsembleMLP` within AutoEmulate, upon `fit_from_reinitialized` an exception is raised.

Thanks @vchhabra-turing for identifying this.

### Reproduction steps

```bash
Running Ensemble MLP:

ae = AutoEmulate(
    x,
    y,
    x_transforms_list=[[]],
    y_transforms_list=[[]],
    models=["ensemblemlp"],
    log_level="progress_bar",
    model_params={"epochs": 500},
)
gp = ae.fit_from_reinitialized(x, y)

Returns error:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 16
      1 x, y = (
      2     torch.tensor(df.iloc[:, :-1].to_numpy()),
      3     torch.tensor(df.iloc[:, -1:].to_numpy()),
      4 )
      7 ae = AutoEmulate(
      8     x,
      9     y,
   (...)     14     model_params={"epochs": 500},
     15 )
---> 16 gp = ae.fit_from_reinitialized(x, y)
     17 # gp = ae.best_result().model

...truncated

TypeError: MLP.__init__() got an unexpected keyword argument 'mlp_kwargs'
Output is truncated. View as a scrollable element or open in a text editor. Adjust cell output settings...
```

### Version

https://github.com/alan-turing-institute/autoemulate/commit/20093a155ffe72bb1165ebce653a2365a79b3937

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### OS

_No response_

sgreenbury:
Closes #914 (merge after #917):

This PR updates the ensemble MLP emulators to have keyword-only arguments (removes `**mlp_kwargs`) due to:
- `**mlp_kwargs` producing exception upon reinitializing
- The alternative keyword arg `mlp_kwargs: dict | None` not being tunable with current approach

It adds a function to allow the generation/extension of MLP subclass docstrings to reduce repetition.

When exploring this, I noticed we currently don't have any `**kwargs` on `Emulator`'s `__init__()` methods - when we run this with `AutoEmulate(..., model_params={"specific_emulator_param": "specific_emulator_param_value"}, ...)` this will raise an exception for models in the compare loop that don't have that keyword argument - supporting this if possible would be worthwhile and might interact with the design choices here.

### DIFF
--- a/autoemulate/emulators/nn/mlp.py
+++ b/autoemulate/emulators/nn/mlp.py
@@ -1,3 +1,5 @@
+import textwrap
+
 from torch import nn
 from torch.optim.lr_scheduler import LRScheduler
 
@@ -18,11 +20,24 @@ def _generate_mlp_docstring(
     Parameters
     ----------
     additional_parameters_docstring: str
-        Subclass specific parameters to include in the docstring.
+        Subclass specific parameters to include in the docstring. This should be passed
+        such as:
+        another_arg_str = '''
+        another_arg: str
+            Description for another_arg.
+        '''
     default_dropout_prob: float | None
         Default value for dropout_prob parameter.
     """
-    return f"""
+    # Unindent the additional parameters docstring
+    additional_parameters_docstring = textwrap.dedent(additional_parameters_docstring)
+    # Re-indent with 4 spaces for docstring formatting
+    if additional_parameters_docstring.strip():
+        additional_parameters_docstring = textwrap.indent(
+            additional_parameters_docstring.strip(), "    "
+        )
+
+    docstring_base = f"""
     Parameters
     ----------
     x: TensorLike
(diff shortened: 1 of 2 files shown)


EdwinB12 (approved):

sgreenbury:
> This looks good to me. My only suggestion is we add a test where the MLP class changes a kwarg value from the default. At the moment, the tests only use test `MLP(x, y)`.

Thanks @EdwinB12 for this point - as discussed it looks like we don't do this systematically across emulators at the moment so we'll open a new issue to address this and check that init args are passed on to instances across all emulators.

> Good point about adding `**kwargs` to the `Emulator`. Alternatively we could explicitly only get the params specific to the model in [here](https://github.com/alan-turing-institute/autoemulate/blob/fa490820a782fae04da2936020e0981f03e91b15/autoemulate/core/compare.py#L458) since we already inspect the `__init__` signature to get default params just above this line.

Thanks @radka-j for identifying this approach - I like the idea of us checking the function signature instead and handling this way as opposed to reintroducing arbitrary `**kwargs` - I'll open an issue for this too.

Issue #906, opened 2025-10-30
Update metrics to be able to access score per task/output


radka-j:
At the moment we average all metrics over tasks. It would be useful to also have access scores per tasks.

radka-j:
Just a note here that I have added a `reduction` kwarg to a new metric in #924. As part of the PR, I also update `evaluate` and `boostrap` functions in `model_selection.py` to accept metric specific kwargs.

PR #924, opened 2025-11-10
Add MSLL metric

### LINKED ISSUE #923: Add mean standardised log loss metric
See equation 2.34 (page 23) [here](https://gaussianprocess.org/gpml/chapters/RW.pdf).

It should look something like:

```python
def msll(y_true, y_pred, y_std):
    first_term = 0.5 * np.log(2 * np.pi * y_std**2)
    second_term = ((y_true - y_pred)**2)/(2 * y_std**2)
    
    return np.mean(first_term + second_term)
```

This can be standardised by subtracting the log loss that would be obtained under the trivial model (i.e., one using the mean and variance of the training data for prediction).

radka-j:
Closes #923

### DIFF
--- a/autoemulate/core/metrics.py
+++ b/autoemulate/core/metrics.py
@@ -373,7 +373,7 @@ class MSLLMetric(ProbabilisticMetric):
                         "distributions."
                     )
                     raise ValueError(msg)
-                return model_nll_output.reshape(*y_true.shape)
+                return model_nll_output.reshape(*y_true.shape[1:])
             msg = (
                 f"Unknown reduction method: {metric_params.reduction}. "
                 "Expected 'mean' or 'none'."
@@ -403,7 +403,7 @@ class MSLLMetric(ProbabilisticMetric):
             if model_nll_output is None:
                 msg = "Per-output MLL not available for non-Independent distributions."
                 raise ValueError(msg)
-            return (model_nll_output - trivial_nll_output).reshape(*y_true.shape)
+            return (model_nll_output - trivial_nll_output).reshape(*y_true.shape[1:])
         msg = (
             f"Unknown reduction method: {metric_params.reduction}. "
             "Expected 'mean' or 'none'."


radka-j:
Following our discussion yesterday @sgreenbury, I think the assumption of MSLL is that you have a Gaussian predictive distribution so I'll update the code to check for that. It doesn't matter for the non-standardised version (MLL) but we also need to update the code to use log_prob of the distribution to handle that case.

autoemulate/core/model_selection.py
```diff
@@ -222,11 +231,17 @@ def bootstrap(
...
         for metric in metrics:
+            kwargs = metric_kwargs.get(metric.name, {})
             metric_scores[metric.name][i] = evaluate(
-                y_pred, y_bootstrap, metric=metric, n_samples=n_samples
+                y_pred,
+                y_bootstrap,
+                metric=metric,
+                n_samples=n_samples,
+                metric_kwargs=kwargs,
             )
```
sgreenbury: I think if going with this approach - I think this might need to be added above for the `n_bootstraps=None` case?

sgreenbury (commented):
I think this is looking good! I left a couple of small comments in addition to the discussion about the `log_prob` (sounds good impling `log_prob` for the test case - an option could also be to directly construct the normal distribution for the training case too and call `log_prob` on this as well for clarity?).

One aspect that could be worth considering revising is the handling of the metric kwargs. Could an alternative be to update the API for the metric call to:

```python
def __call__(self, y_pred, y_true, metric_params: MetricParams | None = None): ...
```
and add a new class for the params, e.g.:
```python
@dataclass
class MetricParams:
    n_samples: int = 1000
    y_train: TensorLike | None = None
    reduction: Literal["mean", "none"] = "mean"
    metric_kwargs: dict | None = None # supports subclasses with arbitrary new kwargs
```
This way all metric params can be captured in one argument and documented in `MetricParams` class. These can also be passed to all instances so we wouldn't need the dictionary lookup for specific metrics (i.e. `y_train` can just be not used within e.g. CRPS, R2, etc). Additionally this could be used in #906. What do you think?

autoemulate/core/metrics.py
```diff
@@ -19,6 +22,30 @@
...
+    Parameters for metric calculations.
+
+    Attributes
+    ----------
+    n_samples: int
+        Number of samples to draw from the predicted distribution if `y_pred` is a
+        distribution. Defaults to 1000.
+    y_train: TensorLike | None
+        Training target values. In MSLL used to parameterize the trivial model for
+        standardization. If None, mean log loss is computed without standardization.
```
sgreenbury:
```suggestion
        standardization. If None, mean log loss is computed without standardization.
        Defaults to None.
```

autoemulate/core/model_selection.py
```diff
@@ -223,10 +228,14 @@ def bootstrap(
...
             metric_scores[metric.name][i] = evaluate(
-                y_pred, y_bootstrap, metric=metric, n_samples=n_samples
+                y_pred,
+                y_bootstrap,
+                metric=metric,
+                metric_params=metric_params,
             )
 
     # Return mean and std for each metric
+
```
sgreenbury:
```suggestion
```

autoemulate/core/metrics.py
```diff
@@ -236,6 +282,134 @@ def __call__(
...
+                    raise ValueError(msg)
+                return model_nll_output
+            msg = (
+                f"Unknown reduction method: {metric_params.reduction}. "
+                "Expected 'mean' or 'none'."
+            )
+            raise ValueError(msg)
+
+        # Ensure 2D y_train for consistent handling
+        y_train_mean = metric_params.y_train.mean(dim=0, keepdim=True).view(1, -1)
```
sgreenbury: I think we can remove restriction to 2D here to support output data that is arbitrary shape e.g. `(B, *output_shape)`
```suggestion
        # Keep original shape for y_train_mean to match y_true shape
        y_train_mean = metric_params.y_train.mean(dim=0, keepdim=True)
```

autoemulate/core/metrics.py
```diff
@@ -236,6 +282,134 @@ def __call__(
...
+
+        if metric_params is None:
+            metric_params = MetricParams()
+
+        # Ensure 2D y_true for consistent handling
+        y_true = y_true.unsqueeze(-1) if y_true.ndim == 1 else y_true
+
+        # Compute mean negative log likelihood (also by output dimension if have
+        # Independent distribution)
+        if isinstance(y_pred, Independent):
```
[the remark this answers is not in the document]
radka-j: Thanks!

PR #956, opened 2026-01-26
Addition of discrepancy GP emulators

cwlanyon:
Discrepancy Emulators are a class of GP emulators that perform transfer learning across a cohort of existing simulators/digital twins. 

Suppose you have a cohort of computational digital twins, $f_1,...,f_N$, where each $f_n$ represents an individual real-world system, and for each $f_n$ you have a Gaussian process emulator, $g_n$. If we want to generate an emulator for a similar system $f_{N+1}$, rather than generating a brand new emulator we can instead emulate the discrepancy between $f_{N+1}$ and the existing cohort.

$$g_{N+1}(\theta) = \sum_{n=1}^N a g_n(\theta) +\delta(\theta)$$

where $a$ are some weights to learn and $\delta$ is the discrepancy, which we model as a GP with its own independent mean and covariance functions. If $f_{N+1}$ is sufficiently similar to the rest of the cohort it should require fewer observations to learn $\delta$ than to learn an independent emulator.

See: https://link.springer.com/article/10.1007/s10439-025-03890-0

To implement Discrepancy GPs in AutoEmulate I have amended autoemulate/autoemulate/emulators/gaussian_process/exact.py to include a DiscrepancyGaussianProcess class and a create_dgp_subclass function to mirror the functionality for creating Gaussian process emulators in AutoEmulate.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -311,6 +311,15 @@
         "doc",
         "code"
       ]
+    },
+    {
+      "login": "z7076",
+      "name": "Jichen Zhang",
+      "avatar_url": "https://avatars.githubusercontent.com/u/69681924?v=4",
+      "profile": "https://github.com/z7076",
+      "contributions": [
+        "code"
+      ]
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -7,12 +7,8 @@
 [![Documentation](https://img.shields.io/badge/documentation-blue)](https://alan-turing-institute.github.io/autoemulate/)
 [![Github Stats](https://img.shields.io/badge/repostats-orange)](https://github.com/alan-turing-institute/autoemulate/blob/github-repo-stats/alan-turing-institute/autoemulate/latest-report/report.pdf)
 
-
 <!-- SPHINX-START -->
 Simulations of physical systems are often slow and need lots of compute, which makes them unpractical for real-world applications like digital twins, or when they have to run thousands of times for sensitivity analyses. The goal of `AutoEmulate` is to make it easy to replace simulations with fast, accurate emulators. To do this, `AutoEmulate` automatically fits and compares various emulators, ranging from simple models like Radial Basis Functions and Second Order Polynomials to more complex models like Support Vector Machines and  Gaussian Processes to find the best emulator for a simulation. 
-
->[!WARNING]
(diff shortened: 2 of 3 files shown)


radka-j:
@cwlanyon thank you so much for opening this PR! I just wanted to let you know that reviewing it is on our TODO list but we are currently working towards a tight deadline on our project so we have not had the capacity. We hope to get to it very soon.

autoemulate/emulators/gaussian_process/exact.py
```diff
@@ -677,3 +677,405 @@ def get_tune_params():
...
+        self.early_stopping = early_stopping
+        self.posterior_predictive = posterior_predictive
+        self.num_tasks = num_tasks
+        self.to(self.device)
+
+
+class aKernel(gpytorch.kernels.Kernel):
+    """
+    Custom kernel for Discrepancy GPs.
+    """
```
radka-j: One-line docstring should fit on one line

```suggestion
    """ Custom kernel for Discrepancy GPs."""
```

autoemulate/emulators/gaussian_process/exact.py
```diff
@@ -677,3 +677,405 @@ def get_tune_params():
...
+        length_constraint=None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        self.a = a
+        self.ref_model = ref_model
+        self.ref_likelihood = ref_likelihood
+
+    def forward(self, x1, x2, **params):
```
radka-j: missing docstring in public method

autoemulate/emulators/gaussian_process/exact.py
```diff
@@ -677,3 +677,405 @@ def get_tune_params():
...
+        batch_shape=torch.Size(),
+        bias=True,
+    ):
+        super().__init__()
+        self.a = a
+        self.ref_model = ref_model
+        self.ref_likelihood = ref_likelihood
+        self.mean_module = mean_module
+
+    def forward(self, x):
```
radka-j: missing docstring in public method

radka-j (commented):
This should be all the pre-commit issues. As you can see, most of it is just enforcing strict stylistic guidelines (e.g., max line lengths, docstring conventions). There is absolutely no rush on addressing them before we do a more thorough review of the whole code but hopefully it's useful in explaining what kinds of things get picked up.

The main thing to go over are the unused method arguments and decide whether these should be removed or whether there are some possible future use cases or compatibility issues with the rest of the code base that mean we should keep them.

autoemulate/emulators/gaussian_process/exact.py
sgreenbury: Are `**params` needed in the signature here - perhaps they can be removed?

PR #963, opened 2026-02-05
Add pyomofy Pyomo conversion support

GhostyWave:
A standalone `pyomofy` function has been added to convert a fitted AutoEmulate model into Pyomo-compatible algebraic expressions, enabling surrogate-based optimization with solvers like Ipopt.

## Location

The function lives in its own module, seperate from any specific emulator class:

```
autoemulate/
  convert/
    pyomo.py   ← pyomofy lives here
```

## Usage

```python
from autoemulate.convert.pyomo import pyomofy

ae = AutoEmulate(...)
model = ae.best_result()
expressions = pyomofy(model, [pyo_model.x1, pyo_model.x2])
model_expr = expressions[0]
```

## API

```python
pyomofy(model: TransformedEmulator | Result, pyomo_vars, relu_beta: int = 50)
```

- `model` — a `TransformedEmulator` or a `Result` object returned by `ae.best_result()`.
- `pyomo_vars` — list of Pyomo decision variables, in the same order as the training inputs.
- `relu_beta` — sharpness parameter for the Softplus approximation used when the model was trained with `nn.ReLU`. Only relevant for MLP models with ReLU activations. Higher values produce a tighter approximation to ReLU; lower values are smoother and more Ipopt-friendly. Defaults to 50.

## Supported Emulators

- `PolynomialRegression`
- `MLP`

## ReLU Handling

`nn.ReLU` is not differentiable at zero, which can cause gradient-based solvers like Ipopt to stall. When `pyomofy` encounters a ReLU layer, it automatically substitutes a Softplus approximation with a tunable sharpness parameter:

$$\text{Softplus}_\beta(x) = \frac{1}{\beta} \log(1 + e^{\beta x})$$

As $\beta \to \infty$, this converges to ReLU. The default `relu_beta=50` provides a close approximation while remaining smooth. Users can override this via the `relu_beta` argument if needed.

## Supported Transforms

Only the built-in AutoEmulate transforms are currently supported:

- `StandardizeTransform`  —  linear, fully expressible in Pyomo.
- `PCATransform` — linear, fully expressible in Pyomo.

`VAETransform` is **not supported** because its encoder/decoder are non-linear neural networks, making the inverse non-differentiable and impractical to embed in a Pyomo model. A `NotImplementedError` is raised if an unsupported transform is detected.

## Supported Activations (MLP)

| Activation | Handling |
|---|---|
| `nn.ReLU` | Approximated via `Softplus(beta)` |
| `nn.SiLU` | Natively supported |
| `nn.Tanh` | Natively supported |
| `nn.Sigmoid` | Natively supported |
| `nn.Softplus` | Natively supported |

Any other activation raises a `TypeError` listing the supported options.

## Documentation Updates

- **Reference docs**
  - `docs/reference/convert/index.md`
  - `docs/reference/convert/pyomo.rst`
- **Tutorial notebook**
  - `docs/tutorials/convert/01_pyomo_conversion.ipynb`
- **Table of contents**
  - `docs/_toc.yml` updated to include the new reference and tutorial pages


GhostyWave:
Hello @radka-j 

I saw that the previous commit had some typesetting issues. I updated the code and have uploaded it.

appreciated, if you could run the checks again, and guide for the next steps ahead.

docs/tutorials/convert/01_pyomo_conversion.ipynb
```diff
@@ -0,0 +1,306 @@
...
+    "x.shape, y.shape"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Data\n",
+    "\n",
+    "As you can see, our simulator inputs (`x`) and outputs (`y`) are PyTorch tensors. PyTorch tensors are a common data structure used in machine learning, and `AutoEmulate` is built to work with them."
```
radka-j: I would take this bit out as we cover elsewhere that AutoEmulate is PyTorch based

docs/tutorials/convert/01_pyomo_conversion.ipynb
```diff
@@ -0,0 +1,306 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from autoemulate import AutoEmulate\n",
+    "\n",
+    "# Initialize AutoEmulate with MLP model only\n",
+    "ae = AutoEmulate(x, y, log_level=\"info\", models=['MLP'])"
```
radka-j: You can just get the trained emulator and not worry too much about explaining this step as it is already covered elsewhere in the docs

```suggestion
    "ae = AutoEmulate(x, y, log_level=\"info\", models=['MLP'])",
    "best = ae.best_result()\n"
```

autoemulate/convert/pyomo.py
```diff
@@ -0,0 +1,326 @@
...
+_SUPPORTED_TRANSFORMS = (StandardizeTransform, PCATransform)
+
+# Default beta for the Softplus approximation when the model was trained with ReLU.
+# As beta -> inf, Softplus_beta -> ReLU. beta=50 is sharp enough to closely track
+# ReLU while remaining smooth and differentiable everywhere.
+_DEFAULT_RELU_BETA = 50
+
+
+def pyomofy(
+    model: TransformedEmulator | Result, pyomo_vars, relu_beta: int = _DEFAULT_RELU_BETA
```
radka-j: You could also let the user pass in an MLP or PolynomialRegression directly (not within TransformedEmulator), in which case it just means that there are no transforms attached. These could be handled by allowing for `PytorchBackend` type and then handling that in the init as well.

radka-j (commented):
The code looks great @GhostyWave! I just had one small comment and question that it would be good to clarify.

autoemulate/convert/pyomo.py
radka-j: I am happy to leave the code as is (i.e., user passes in `Results` or `TransformedEmulator` object!

It did make me realise that the underlying emulator model (e.g., MLP) also has a `x_transforms` and `y_transforms` attributes (this can be set to `StandardizeTransform`). Should we check whether these transforms have been set and then handle them the same way as when they are attached to the `TransformedEmulator`?

radka-j: Yes exactly. It's another layer so it should be considered. It should be rare that transforms are defined within the base model itself (e.g., `MLP`) as well within the `TransformedEmulator` but it is possible. If it is within the base model, then it is just a transform applied to the data (there is no inverse transform).

Issue #978, opened 2026-04-13
test failures and warnings


pmeier:
_Disclaimer: I'm reviewing the `autoemulate` package as part of this [JOSS review](https://github.com/openjournals/joss-reviews/issues/10087). This issue is based on version [`v1.1.1.`](https://github.com/alan-turing-institute/autoemulate/tree/v1.1.1)_

---

`uv run pytest` results in 16 failures and 324 warnings:

```
============================================================= warnings summary ==============================================================
.venv/lib/python3.12/site-packages/pyro/ops/stats.py:514
  /home/philip/git/github.com/autoemulate/.venv/lib/python3.12/site-packages/pyro/ops/stats.py:514: SyntaxWarning: invalid escape sequence '\g'
    """

tests/calibration/test_bayesian_calibration.py::test_hmc_single_output[1-1-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_single_output[10-1-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_single_output[1-2-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_single_output[10-2-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_multiple_output[1-1-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_multiple_output[10-1-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_multiple_output[1-2-10-False]
tests/calibration/test_bayesian_calibration.py::test_hmc_multiple_output[10-2-10-False]
  /home/philip/git/github.com/autoemulate/autoemulate/calibration/bayes.py:189: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.detach().clone() or sourceTensor.detach().clone().requires_grad_(True), rather than torch.tensor(sourceTensor).
    else torch.tensor(observation_variance + self.model_discrepancy).to(

tests/core/test_compare.py: 3 warnings
tests/emulators/test_transformed.py: 90 warnings
  /home/philip/git/github.com/autoemulate/autoemulate/emulators/ensemble.py:291: NumericalWarning: cov not p.d. - added 1.0e-04 to the diagonal and symmetrized
    make_positive_definite(

tests/core/test_plotting.py::test_display_figure_jupyter
tests/core/test_plotting.py::test_display_figure_terminal
[...]

PR #983, opened 2026-04-14
Add method to move TransformedEmulator to new device

radka-j:
Contributes to #978 

This PR adds a `TransformedEmulator.to(device)` method.

In the calibration workflows we let users define a new device, which might differ from the one that the emulator was trained on. Irrespective, we should support training an emulator on one device and then moving it to another. This is a bit more involved for the TransformedEmulator as we have to handle the different types of emulators as well as all the transforms and associated data. This is the motivation for adding this method. 

This PR resolves a test failure we were seeing with a device mismatch in history matching.


sgreenbury (commented):
Looks great - I've added some refactors/refinements in #986 for this to generalize the idea across all the classes with state on devices.

sgreenbury (approved):
LGTM!

PR #985, opened 2026-04-14
Prevent CUDA test failures from deterministic algorithm mode

radka-j:
Contributes to #978 

On CUDA, some operations don't have deterministic implementations and this was causing tests to fail. This PR makes the following changes:
- use `warn_only=True` when using deterministic algorithms to avoid hard failures and we 
- set `deterministic=False` by default when using `set_random_seed` (this is consistent with default PyTorch behaviour which also has this as opt in) 
- add `deterministic` arg option to all classes and methods that call `set_random_seed` so that the user can reset it
-  add a fixture to conftest to reset global PyTorch deterministic settings back to defaults after each test


autoemulate/experimental/emulators/nn/gaussian_mlp.py
```diff
@@ -198,3 +201,4 @@ def _predict(self, x: TensorLike, with_grad: bool) -> GaussianLike:
     def loss_func(self, y_pred, y_true):
         """Negative log likelihood loss function."""
         return -y_pred.log_prob(y_true).mean()
+        return -y_pred.log_prob(y_true).mean()
```
radka-j:
```suggestion
```

sgreenbury (approved):
Looks great - thanks @radka-j! Only thought was that we could extend this to other emulator classes where `random_seed` is passed: https://github.com/alan-turing-institute/autoemulate/pull/987

PR #988, opened 2026-05-07
Refactor logging

### LINKED ISSUE #980: logging should not be configured
_Disclaimer: I'm reviewing the `autoemulate` package as part of this [JOSS review](https://github.com/openjournals/joss-reviews/issues/10087). This issue is based on version [`v1.1.1.`](https://github.com/alan-turing-institute/autoemulate/tree/v1.1.1)_

---

Quoting the [official Python documentation](https://docs.python.org/3/howto/logging.html#library-config):

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

Yet

https://github.com/alan-turing-institute/autoemulate/blob/24f32d7fa0cc9d167fab3cb1ed1570b85ca7f799/autoemulate/core/logging_config.py#L7-L21

radka-j:
Closes #980 

This pull request refactors logging to standardize usage and simplify configuration. It replaces custom logger setup and progress bar logic with a unified `get_logger` approach, deprecates the `log_level` argument, and updates docstrings and internal logic accordingly. Additionally, it introduces a `show_progress_bar` argument in `AutoEmulate` for progress display, decoupling it from logging configuration. 

Logger is initialised at package import by calling `get_logger()` in `autoemulate/__init__.py` . We use `get_logger(__name__)` so that logs are named by the module they belong to.

These changes make logging more consistent and easier to configure, while simplifying the API for users and developers.


autoemulate/core/compare.py
```diff
@@ -120,14 +122,10 @@ def __init__(
...
+            Whether to show a progress bar during model comparison. Defaults to True.
         deterministic: bool
-            Whether to use deterministic algorithms in PyTorch. Defaults to False.
-        log_level: str
-            Logging level. Can be "progress_bar", "debug", "info", "warning",
-            "error", or "critical". Defaults to "progress_bar". If "progress_bar",
-            it will show a progress bar during model comparison. It will set the
-            logging level to "error" to avoid cluttering the output
-            with debug/info logs.
+            Whether to use deterministic algorithms in PyTorch. Defaults to False..
```
penelopeysm:
```suggestion
            Whether to use deterministic algorithms in PyTorch. Defaults to False.
```

penelopeysm (commented):
Not sure if anybody asked 😅 but looks reasonable to me.

> while simplifying the API for users and developers.

Could I ask how versioning has been handled to date? Specifically: have autoemulate releases obeyed semver (for example technically this is a breaking change to the API so to be semver compliant this would have to be bundled into a 3.0 release)? Are there changelogs somewhere to help users with upgrading?

autoemulate/core/logging_config.py
```diff
@@ -1,125 +1,36 @@
...
-
-    warnings_logger = logging.getLogger("py.warnings")
-    for handler in logger.handlers:
-        warnings_logger.addHandler(handler)
-    warnings_logger.setLevel(logger.getEffectiveLevel())
 
+def _setup_library_logger() -> logging.Logger:
+    """Ensure the package root logger has a NullHandler for library-safe logging."""
+    logger = logging.getLogger(PACKAGE_LOGGER_NAME)
+    if not any(isinstance(handler, logging.NullHandler) for handler in logger.handlers):
```
penelopeysm:
```suggestion
    if len(logger.handlers) == 0:
```

I think it is probably more correct to only attach a NullHandler if there are no handlers. If there are already (other) handlers, then adding a NullHandler is redundant.

(I prefer the explicit length check but you can do `if not logger.handlers` too, of course)

autoemulate/simulations/base.py
```diff
@@ -57,7 +50,8 @@ def __init__(
         self._in_dim = len(self.param_names)
         self._out_dim = len(self.output_names)
         self._has_sample_forward = False
-        self.logger, self.progress_bar = get_configured_logger(log_level)
+        self.logger = get_logger(__name__)
```
penelopeysm: Is there a reason to keep using `self.logger` here as opposed to the module-level `logger`?

autoemulate/__init__.py
```diff
@@ -1,3 +1,6 @@
 from .core.compare import AutoEmulate
+from .core.logging_config import get_logger
+
+get_logger()
```
penelopeysm, line 4: This call to `get_logger()` calls `_setup_library_logger()` inside `core/logging_config.py`, but the last line of that file also already calls `_setup_library_logger()` at the top level. Probably it doesn't really hurt to call it multiple times, but I think we could remove one?

sgreenbury:
> Could I ask how versioning has been handled to date? Specifically: have autoemulate releases obeyed semver (for example technically this is a breaking change to the API so to be semver compliant this would have to be bundled into a 3.0 release)? Are there changelogs somewhere to help users with upgrading?

Good point re. the breaking change. Following the bump to v2.0 we're aiming to follow semver. An option here for this to be non-breaking could be to:
- Retain the current function API and importable functions
- Add `show_progress_bar` last positionally in the args
- Emit a deprecation warning where values are passed that use the deprecated API
- Update the docstrings to indicate that `log_level` is deprecated

radka-j:
Thank you for the review @penelopeysm, really appreciate it!!

autoemulate/__init__.py
radka-j: Good spot!

autoemulate/simulations/base.py
radka-j: Good spot, we should refactor so that this is no longer needed.

sgreenbury (approved):
LGTM!

PR #987, opened 2026-05-07
Add deterministic flag for wrapping classes

sgreenbury:
This PR:
- Forwards the deterministic kwarg through EnsembleMLP, EnsembleMLPDropout, ZOIBMLP, and AutoEmulate.fit_from_reinitialized so users instantiating these directly (without going through AutoEmulate) can opt into torch.use_deterministic_algorithms via the same API as MLP.
- Fixes same random seed passed to EnsembleMLP

PR #996, opened 2026-05-12
Fix deterministic argument order

sgreenbury:
This PR revises the deterministic argument added in #987 to ensure the change is non-breaking.


radka-j (approved):
LGTM!

PR #997, opened 2026-05-12
Remove warning suppression from docs

sgreenbury:
This PR removes the suppression of warnings from docs (notebooks and script).


radka-j (approved):
LGTM!

PR #1006, opened 2026-05-14
Bump sklearn compat

### LINKED ISSUE #991: Relax scikit-learn bounds
scikit-learn is pinned to `>=1.3, <1.6` by [this commit](https://github.com/alan-turing-institute/autoemulate/commit/d3e9a76df4f66e2adb7d1827f3f542957ea01083) which doesn't explain much. Is there still a reason for this? It seems a bit restrictive, see e.g. https://github.com/alan-turing-institute/autocast/issues/360.

Claude says

> sklearn.utils.validation.check_X_y was deprecated in scikit-learn 1.6 (replaced by sklearn.utils.validation.validate_data). So when 1.6 came out in December 2024, it probably threw deprecation warnings or broke something, and the quick fix was to cap at <1.6.0. There's no deep incompatibility — it's just check_X_y usage in svm.py and data/utils.py. Replacing those calls with the new API (or just using validate_data on the estimator) would let you lift the upper bound. It's a small fix that would unblock users who have newer sklearn installed.

but I haven't verified if this is correct.

penelopeysm:
Closes #991. I recognise that it's not mandatory to have autocast and autosim in the same environment since it is quite possible (and indeed maybe even preferred) to run them separately and communicate via the data files, but I think we should at least not _forbid_ people from doing that if they need or want to.


pyproject.toml
```diff
@@ -18,7 +18,7 @@ classifiers = [
 dependencies = [
     "matplotlib>=3.7.2",
     "seaborn>=0.12.2",
-    "scikit-learn>=1.3.0,<1.6.0",
+    "scikit-learn>=1.3.0",
```
penelopeysm, line 21: Tested locally with sklearn 1.7 and 1.8 and didn't get any errors/depwarns.

I suspect that there is code inside the `v0/` directory that might error with newer sklearn, but I don't think any of that is tested. Is that code meant to be part of the public API or is it just "here's legacy code that may or may not work"?

(If the latter, it should arguably be removed from the repo, but that's a separate thing)

pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
     "torchmetrics>=1.7.1",
     "pyro-ppl>=1.9.1",
     "torchrbf>=0.0.1",
-    "arviz>=0.21.0",
+    "arviz>=0.21.0,<1.0",
```
penelopeysm, line 37: Had to pin this because arviz 1.0 changes the signature of `arviz.from_dict()` which is used in `calibration/interval_excursion_set.py`

uv.lock
penelopeysm, line 1: I went ahead and upgraded the deps in here, please tell me if that's not the right approach

penelopeysm: ~~CI failures are caused by upgrading torch here from 2.7 to 2.12~~

Resolved by not upgrading anything in the lockfile apart from sklearn

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "autoemulate"
-version = "2.1.1"
+version = "2.1.2"
```
sgreenbury: Perhaps we should bump this just at the release time in a distinct PR
```suggestion
version = "2.1.1"
```

pyproject.toml
sgreenbury: Good point, yes that's the legacy pre-v1.0 we retained for visibility, perhaps this can now be removed in a separate PR.

sgreenbury (approved):
Looks great, thanks @penelopeysm! I just left one suggestion below and I also think it might be worth adding an ignore to the [`pyproject.toml`](https://github.com/alan-turing-institute/autoemulate/blob/87f3a4bd558fcfc6f669a5732b5a05c9c77540e2/pyproject.toml#L118) for the following warning that's raised when running the tests with this updated version:
```toml
filterwarnings = [
    # ...
    "ignore:X does not have valid feature names, but LGBMRegressor was fitted with feature names:UserWarning",
]
```
Approving as looks good to merge otherwise!

pyproject.toml
penelopeysm: Sure

pyproject.toml
```diff
@@ -125,6 +125,10 @@ filterwarnings = [
     "ignore:jax\\.interpreters\\.xla\\.pytype_aval_mappings is deprecated\\..*:DeprecationWarning:tensorflow_probability\\.python\\.internal\\.backend\\.jax\\.ops",
     # PyTorch MPS backend capability fallbacks are backend noise in tests.
     "ignore:The operator .* is not currently supported on the MPS backend.*:UserWarning",
+    # LightGBM automatically adds feature names to training data, which causes warnings when
+    # making predictions with unnamed data
+    # https://github.com/lightgbm-org/LightGBM/issues/6798
+    "ignore:X does not have valid feature names, but LGBMRegressor was fitted with feature names:UserWarning",
```
penelopeysm, line 131: @sgreenbury Thanks! I went down a bit of a rabbit hole trying to figure out where this happened and whether we could work around it in code rather than suppressing the warning, and it's a known issue, so just suppressing it makes sense.

More broadly, if we'd like to make sure that CI doesn't show warnings, would it make sense to promote all (unfiltered) warnings to errors? We could do that by adding `"error"` as the first entry to this list (reference more for myself than anything, https://docs.pytest.org/en/stable/reference/reference.html#confval-filterwarnings)

penelopeysm: (Looking at the CI log I think there's only one other warning apart from this, so I could try to squish that in a separate PR.)

sgreenbury: Thanks for investigating this further and the info! Great idea for promoting the non-filtered warnings, that makes sense - let's open an issue for this.

penelopeysm: OK, will do! And will merge this when CI passes, thanks :)

sgreenbury: Thanks!

PR #1013, opened 2026-06-03
Add SIR vs SEIR Bayes factor model comparison notebook to case studies

SyedHaider2084:


radka-j (commented):
We should break everything up into sections and add text explaining what this notebook is doing - the goal here is to introduce model comparison. 

The notebook should walk the user through everything step by step. Have a look at the patient calibration case study - that one is more involved but gives you the idea. 

The plan is to remove the original SEIR notebook that was in docs so some of that content can be moved here and then add content about model comparison and how we use evidence here. @z7076 can then add a section comparing the results here to nested sampling at the end.

radka-j:
@SyedHaider2084 I don't see any changes here, I think you didn't push them?

radka-j:
@all-contributors please add @SyedHaider2084 for docs

radka-j:
@all-contributors please add @z7076 for doc

radka-j:
@SyedHaider2084 this looks really great!!! I only have a couple of minor comments:

Notebook:
- it's probably worth acknowledging somewhere at the start of the notebook that the SIR/SEIR comparison is a toy problem (we don't really need an emulator) but that the workflow would be the same for any other model and that is the goal here
- in section 2, 1500 training data points might be excessive for such a simple problem (also, we do 2000 evaluations in the MCMC, you could just as well use the simulator at this point without training an emulator)
- it might be nice to add a link to the harmonic package where we introduce it
- at the end of section 4 presumably the evidence is not in favour of either model, this is worth acknowledging (and here we can call back to the start if we pointed out that we are looking at a toy problem and there is not enough to distinguish the two models)
- similarly, it would be nice to add a link to dynesty docs at the top of section 5 for more information on nested sampling and implementation

Case studies structure:
- each case study should be in its own directory (i.e., something like `autoemulate/case_studies/model_comparison/sir_seir.ipynb`
- the notebook should be accompanied with a `pyproject.toml` file that lists all dependencies (in this case that's `autoemulate` and `dynesty`)
- there can also be a short README in the `model_comparison` directory giving a quick overview of what it is and how to run it
- a short overview should be added to the [README](https://github.com/SyedHaider2084/autoemulate/blob/ea6e9687954090be469a29440adc534899ca8f2d/case_studies/README.md) in the `case_studies` directory

case_studies/.ipynb_checkpoints/sir_seir_model_comparison-checkpoint.ipynb
```diff
@@ -0,0 +1,408 @@
+{
```
radka-j: Delete this file (and the whole .ipynb_checkpoints directory)

case_studies/model_comparison/sir_seir.ipynb
```diff
@@ -0,0 +1,322 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "33560abc-f0f5-4a04-8b79-58b7791041d2",
+   "metadata": {},
+   "source": [
+    "## 5. Summary and Results\n",
+    "\n",
+    "*Note: For more information on alternative robust evidence computation methods like Nested Sampling, you can refer to the [`dynesty` documentation](https://dynesty.readthedocs.io/).*\n",
```
radka-j: What happened to the nested sampling section?

case_studies/README.md
```diff
@@ -7,6 +7,7 @@ This directory contains example case studies demonstrating the use of the AutoEm
 Each subdirectory provides data, scripts, and documentation relevant to a specific application area.
 
 - **patient_calibration/**: Demonstrates cardiovascular model calibration using the Naghavi Model from [ModularCirc](https://github.com/alan-turing-institute/ModularCirc). Shows end-to-end workflow including sensitivity analysis, history matching, and Bayesian calibration for blood pressure dynamics in the left ventricle.
+- **model_comparison/**: A toy problem case study demonstrating how to perform surrogate calibration and compute Bayesian Evidence. It uses Gaussian Process emulators to quantitatively compare SIR and SEIR epidemic models and introduces the concept of the Bayes Factor to guard against overfitting.
```
radka-j:
```suggestion
- **model_comparison/**: Demonstrates Bayesian model comparison using SIR and SEIR epidemic models to determine which model best explains observed data. Shows how to perform Bayesian calibration,  compute model evidence and compare models using Bayes factor.
```

case_studies/model_comparison/sir_seir.ipynb
```diff
@@ -0,0 +1,322 @@
...
+   "source": [
+    "# Bayesian Model Comparison: SIR vs. SEIR\n",
+    "\n",
+    "## Introduction\n",
+    "When dealing with complex physical or biological systems, we often have competing mathematical models that explain the same phenomena. A critical question arises: **which model best explains our observed data without unnecessarily overfitting?**\n",
+    "\n",
+    "This case study introduces **Bayesian Model Comparison**. We will evaluate two competing epidemiological models—the standard SIR model and the extended SEIR model—against a single set of observed data. \n",
+    "\n",
+    "### A Note on this Case Study\n",
+    "It is worth acknowledging that the SIR and SEIR models used here are simple \"toy\" problems. Because integrating these specific ODEs is very fast, we wouldn't strictly *need* an emulator to perform Bayesian calibration here. However, the purpose of this notebook is to demonstrate a highly scalable **surrogate calibration workflow**. The pipeline outlined below is exactly what you would use for complex, computationally expensive simulators where direct Markov Chain Monte Carlo (MCMC) evaluation would take days or weeks.\n",
```
radka-j:
```suggestion
    "It is worth acknowledging that the SIR and SEIR models used here are simple \"toy\" problems. Because integrating these specific ODEs is very fast, we wouldn't strictly *need* an emulator to perform Bayesian calibration here. However, the pipeline outlined below is exactly what you would use for complex, computationally expensive simulators .\n",
```

case_studies/model_comparison/sir_seir.ipynb
```diff
@@ -0,0 +1,322 @@
...
+   "source": [
+    "## 5. Summary and Results\n",
+    "\n",
+    "*Note: For more information on alternative robust evidence computation methods like Nested Sampling, you can refer to the [`dynesty` documentation](https://dynesty.readthedocs.io/).*\n",
+    "\n",
+    "Below, we output the posterior summaries for both models alongside the computed log-evidence and the Bayes Factor. \n",
+    "\n",
+    "**Interpreting the Results:** Because our ground truth data was generated with an incubation rate ($\\sigma = 0.2$), you might expect a strong preference for the SEIR model. However, because this is a toy problem relying strictly on a single data point type (peak infection rate), there is not enough information to break the geometric degeneracy between the parameters. Consequently, you will likely see a Bayes Factor indicating only **weak evidence** in favor of either model. \n",
+    "\n",
+    "This beautifully illustrates the concept of Bayesian model comparison: the evidence inherently penalizes the SEIR model's extra complexity because the simple peak infection data is not rich enough to justify the additional parameter over the prior volume."
```
radka-j:
```suggestion

```

radka-j (commented):
It's almost there, I only made some very minor wording tweak suggestions. 

My main question is about nested sampling - I thought we had a comparison to this method in the notebook? Now there is just a link to docs of a nested sampling package. If we had the comparison, we should add it back in. If not, we should remove the package from the pyproject.toml.

radka-j (approved):

PR #1029, opened 2026-07-07
Fix #928: Replace ±2σ with 95% predictive intervals in plots

### LINKED ISSUE #928: Update plots to display 95% predictive intervals instead of +/- 2sigma
Some methods, like conformal prediction, only give us the interval rather than sigma.

hebatea:
Replaces 2 * y_std (±2σ) with 1.96 * y_std (95% PI) in all plots and updates labels accordingly.

- Adds PREDICTION_INTERVAL_Z = 1.96 constant in plotting.py and v0/plotting.py so future coverage changes are a single-line edit
- Updates plot_xy (bars + fill) and plot_preds in compare.py
- Updates docstrings
- Adds 3 tests verifying the interval width

### DIFF
--- a/autoemulate/core/plotting.py
+++ b/autoemulate/core/plotting.py
@@ -84,7 +84,7 @@ def plot_xy(
         An array of predictions.
     y_variance: NumpyLike | None
         An optional array of predictive variances. When provided, a 95%
-        predictive interval (mean ± 1.96σ) is displayed around predictions.
+        predictive interval (mean +/- 1.96 * std) is displayed around predictions.
     ax: Axes | None
         An optional matplotlib Axes object to plot on.
     title: str
@@ -196,7 +196,7 @@ def plot_xy(
     # Place R² just below the legend
     if legend:
         # Get the bounding box of the legend in axes coordinates
-        bbox = legend.get_window_extent(ax.figure.canvas.get_renderer())  # pyright: ignore[reportAttributeAccessIssue]
+        bbox = legend.get_window_extent(ax.figure.canvas.get_renderer())  # pyright: ignore[reportAttributeAccessIssue, reportOptionalMemberAccess]
         inv = ax.transAxes.inverted()
         bbox_axes = bbox.transformed(inv)
         # Place the text just below the legend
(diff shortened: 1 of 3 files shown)


hebatea:
@radka-j  could you review this PR? It fixes #928 by replacing ±2σ with 95% predictive intervals in plots.

hebatea:
@radka-j Fixed the failing test issue and pushed the updates. Ready for review again,

tests/core/test_plotting.py
```diff
@@ -53,6 +53,41 @@ def test_plot_xy():
     assert len(ax.collections) > 0
 
 
+def test_prediction_interval_z_constant():
+    assert plotting.PREDICTION_INTERVAL_Z == 1.96
```
penelopeysm: The comparison against a literal isn't really meaningful in itself. The test is checking that the constant has the right value, but by defining the correct value as a literal in this test, the logic becomes circular i.e. how do you know that your test is testing the right thing?

One improvement would be to test that `PREDICTION_INTERVAL_Z * stdev` does in fact yield a 95% interval (i.e., test that `P(µ - PREDICTION_INTERVAL_Z*σ ≤ x ≤ µ + PREDICTION_INTERVAL_Z*σ) = 0.95`).

However, I think it would be best to just define PREDICTION_INTERVAL_Z as not a hardcoded literal but rather by `scipy.stats.norm.ppf`. Then you don't even need to test it, because by definition it is correct. On top of that, it's more accurate than 1.96, which is slightly approximated (although I recognise the difference is very small).

autoemulate/core/plotting.py
```diff
@@ -13,6 +13,9 @@
 
 _NON_INTERACTIVE_BACKENDS = {"agg", "cairo", "pdf", "pgf", "ps", "svg", "template"}
 
+# z-score for the default predictive interval coverage (norm.ppf(0.975) ≈ 1.96 for 95%)
+PREDICTION_INTERVAL_Z = 1.96
```
penelopeysm:
```suggestion
# z-score for the default predictive interval coverage (≈ 1.96 for 95% interval)
PREDICTION_INTERVAL_Z = norm.ppf(0.975)
```

autoemulate/core/plotting.py
penelopeysm, line 9:
```suggestion
from matplotlib.figure import Figure
from scipy.stats import norm
```

penelopeysm (changes requested):

penelopeysm (approved):
Looks great :)

radka-j:
@all-contributors please add @hebatea for code

PR #1033, opened 2026-08-03
Fix docstring parameters that do not match the signatures

karpovantonme:
## What this fixes

Seven docstring blocks across five files name a parameter the function does not take, or leave a real one undocumented. All of them are reachable from the public API, so copying the documented name gives a `TypeError`.

| Where | Documented | Actual signature |
|---|---|---|
| `BayesianCalibration.__init__` | `parameters_range` (×3, incl. two mentions in Notes) | `parameter_range` |
| `Results.best_result` | `metric_maximize` | no such parameter |
| `GaussianProcessCorrelated.__init__` | `activation` | no such parameter |
| `GaussianProcess.__init__`, `GaussianProcessCorrelated.__init__` | — | `standardize_x`, `standardize_y` undocumented |
| `GaussianProcessCorrelated.__init__` | `scheduler_cls` description ends at "Defaults to" | base class has the same text ending "None." |
| `_plot_sa_heatmap` | `results`, `top_n`, `figsize` | `si_df`, `parameters`, `fig_size` |
| `Learner.query` | `*arg` | `x` |

A few notes on the individual ones:

**`parameters_range` vs `parameter_range`.** Both spellings genuinely exist in the codebase — `core/plotting.py` and `core/compare.py` do take `parameters_range`, which is probably where the docstring came from. Only `BayesianCalibration.__init__` is inconsistent with its own signature (`self.parameter_range = parameter_range`, and `calibration_params` defaults to `parameter_range.keys()`), so I changed the docstring rather than the argument.

**`metric_maximize`.** The behaviour it describes moved into the metric object — `best_result` branches on `metric_selected.maximize`. I folded that into the `metric` description instead of just deleting the block.

**`activation`.** No `nn.Module` activation is constructed anywhere in `GaussianProcessCorrelated`; the mean and covariance modules come from `mean_module_fn` / `covar_module_fn`.

**`standardize_x` / `standardize_y`.** Real arguments in both constructors (`self.x_transform = StandardizeTransform() if standardize_x else None`), with different defaults from each other (`False` and `True`), and undocumented in both. Wording taken from `StandardizeTransform`.

Docstrings only — no behaviour change, nothing outside docstrings touched. `ruff format --check` and `ruff check` are clean on the five files.

I did not touch the `__post_init__` docstrings in `learners/stream.py`, which document `x_train` / `y_train` as dataclass fields rather than arguments — that reads deliberate, but say the word if you would like them aligned too.

### DIFF
--- a/autoemulate/calibration/bayes.py
+++ b/autoemulate/calibration/bayes.py
@@ -42,7 +42,7 @@ class BayesianCalibration(TorchDeviceMixin, BayesianMixin):
         ----------
         emulator: Emulator
             Fitted Emulator object.
-        parameters_range : dict[str, tuple[float, float]]
+        parameter_range : dict[str, tuple[float, float]]
             A dictionary mapping input parameter names to their (min, max) ranges.
         observations: dict[str, TensorLike]
             A dictionary of observations for each output.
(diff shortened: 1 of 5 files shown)


radka-j:
@karpovantonme thanks for spotting and fixing these, very helpful!

radka-j:
@all-contributors please add @karpovantonme for docs

radka-j (approved):

PR #1046, opened 2026-09-03
Fix harmonic install issues

radka-j:
This pull request introduces a build constraint to address a compatibility issue with the `harmonic` package and the latest version of `Cython`. Specifically, it pins the build-time version of `Cython` to below 3.3 to prevent build failures when installing `harmonic` on Python versions earlier than 3.11.

Dependency management:

* Added `build-constraint-dependencies = ["cython<3.3"]` to `pyproject.toml` to avoid build errors with `harmonic` when it is installed from source on Python <3.11, due to a breaking change in `Cython` 3.3.0.

This issue showed up in #1045.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,13 @@ autoemulate = ["**/__pycache__", "**/*.pyc"]
 # so skipping it on Windows is safe. On Linux/macOS it is still installed normally.
 override-dependencies = ["uvloop>=0.0.0; sys_platform != 'win32'"]
 
+# harmonic==1.3.0 (used on Python <3.11, where it has no prebuilt wheel and must
+# be built from source) declares an unbounded `Cython` build requirement. Cython
+# 3.3.0 made `**` on C long types infer a `double` result, which breaks the
+# implicit narrowing assignment in harmonic's model_legacy.pyx and fails the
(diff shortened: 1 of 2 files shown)


sgreenbury (approved):
LGTM!

sgreenbury:
Just to add here that this [came up in AutoCast](https://github.com/alan-turing-institute/autocast/actions/runs/34381169290/job/102566104606) too: the uv build constraint needs repeating in consuming projects when Harmonic is built from source. Since we only needed simulators now available in AutoSim, I removed AutoEmulate in [autocast#453](https://github.com/alan-turing-institute/autocast/pull/453).

Perhaps it would be worth us opening a PR upstream in Harmonic, since this can affect both direct and transitive consumers whenever Harmonic is built from source with an incompatible Cython version?

A further option for AutoEmulate could be e.g. an `autoemulate[evidence]` extra, making Harmonic opt-in. Consumers that don't need evidence estimation could then avoid its dependency stack and the downstream build constraint.

radka-j:
I like the idea of making the harmonic install opt in. Maybe we could also open an issue on their repo asking to fix this?

sgreenbury:
Yep that combination sounds good to me.

